PR #154, opened 2019-08-28
Don't derive port offset from instance name

rosik:


rosik:
@knazarov rejected, closing

PR #157, opened 2019-08-28
Split options http_port and http_enabled

rosik:


CHANGELOG.md
```diff
@@ -59,6 +59,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   `TARANTOOL_INSTANCE_NAME`: `<PORT> = 3300+<N>`.
   Otherwise default `<PORT> = 3301` is used.
 
+- Parameter `http_port` is derived from instance name to. If it can't be derived
```
sharonovd: `from instance name to` ?

vasiliy-t:
this patch introduces confusion in ops ux. The very same could be achieved via httpd user role and it's config

CHANGELOG.md
rosik: Too. It's a typo

PR #158, opened 2019-08-29
new example: cartridge

Alex-Kuz:


examples/simple-project/.cartridge.ignore
```diff
@@ -0,0 +1,4 @@
+# simple cartridge-ignore file
```
sharonovd: I suggest renaming it to getting-started-app

PR #162, opened 2019-08-29
Don't freeze main fiber in cartridge.cfg

printercu:
Release main fiber if remote configuration is not available
immediately.

### DIFF
--- a/cartridge/bootstrap.lua
+++ b/cartridge/bootstrap.lua
@@ -273,6 +273,35 @@ local function bootstrap_from_membership()
     return confapplier.commit_2pc()
 end
 
+local function apply_remote_conf(conf, remote_conf)
+    membership.set_payload('warning', nil)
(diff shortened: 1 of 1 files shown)


cartridge/bootstrap.lua
```diff
@@ -334,40 +363,19 @@ local function bootstrap_from_snapshot()
...
+    local remote_conf = confapplier.fetch_from_membership(conf.topology)
+    if remote_conf == nil then
+        fiber.create(function()
+            while remote_conf == nil do
+                membership.set_payload('warning', 'Configuration is being verified')
+                fiber.sleep(1)
+                remote_conf = confapplier.fetch_from_membership(conf.topology)
+            end
+            apply_remote_conf(conf, remote_conf)
+        end)
```
rosik: return `true`

PR #181, opened 2019-09-03
Implement patch_topology API

rosik:
New API `patch_topology` suitable for editing multiple servers/replicasets
  at once. It can be used for bootstrapping cluster from scratch,
  joining a server to an existing replicaset, creating new replicaset with
  one or more servers, editing uri/labels of servers,
  disabling or expelling servers.


rosik:
Closing in favor of #200

PR #187, opened 2019-09-05
Don't override user-defined vinyl_dir

rosik:


rosik:
This is to be done more gently in context of #189. Closing this PR for now

PR #224, opened 2019-09-26
Public API for checking http auth

### LINKED ISSUE #219: Cartridge should handle http request auth by itself
Currently application authors have to manually check user authorization on all incoming requests. This is not convenient, because they will have to handle many special cases, like cookie expiration. 

Instead of this, I propose that cartridge handles incoming http requests itself, checks if the user is authorized, and then calls any backend methods.

rosik:
New Lua API `cartridge.http_authorize_request()` suitable for checking
  HTTP request headers.

New Lua API `cartridge.http_render_response()` for generating HTTP
  response with proper `Set-Cookie` headers.

New Lua API `cartridge.http_get_username()` to check authorization of
  active HTTP session.

Close #219

### DIFF
--- a/cartridge/auth.lua
+++ b/cartridge/auth.lua
@@ -312,8 +312,7 @@ end
 -- @within Authorizarion
 -- @treturn string|nil
 local function get_session_username()
-    local http_session = fiber.self().storage['http_session']
-    return http_session and http_session.username
+    return fiber.self().storage['http_session_username']
 end
 
 local function login(req)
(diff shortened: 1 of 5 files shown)


cartridge/auth.lua
```diff
@@ -615,42 +621,19 @@ local function remove_user(username)
...
-    local fiber_storage = fiber.self().storage
-    -- clean fiber storage to behave correctly
-    -- when user logouts within keepalive session
-    fiber_storage['auth_session_username'] = nil
-
-    local resp = setmetatable({}, _response_mt)
+local function authorize_request(req)
+    checks('table')
+    local http_session = {}
+    fiber.self().storage['http_session'] = http_session
```
sharonovd: I suggest to hard-wipe fiber.storage for all requrests, via
httpd.before_dispatch hook

PR #240, opened 2019-10-02
New pool.map_call to gather netbox call results

rosik:
```lua

--- Perform a remote call to multiple URIs and map results.
--
-- @function map_call
--
-- @tparam string fn_name
-- @tparam ?table args
-- @tparam[opt] table opts
-- @tparam {string,...} opts.uri_list
--   array of URIs for performing remote call
-- @tparam ?number opts.timeout
--   passed to `net.box` `conn:call` options
--
-- @treturn {URI=value,...}
--   Call results mapping for every URI.
-- @treturn nil|{URI=error,...}
--   Errors mapping for every URI.

local function map_call(fn_name, args, opts)
    checks('string', '?table', {
        uri_list = 'table',
        timeout = '?', -- for net.box.call
    })
end
```

### DIFF
--- a/cartridge/pool.lua
+++ b/cartridge/pool.lua
@@ -17,7 +17,8 @@ local cluster_cookie = require('cartridge.cluster-cookie')
 
 vars:new('locks', {})
 vars:new('connections', {})
-local e_connect = errors.new_class('Pool connect failed')
+errors.new_class('NetboxConnectError')
+errors.new_class('NetboxMapCallError')
 
 --- Enrich URI with credentials.
 -- Suitable to connect other cluster instances.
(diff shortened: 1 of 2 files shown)


cartridge/pool.lua
```diff
@@ -69,13 +83,82 @@ local function connect(uri, options)
...
+            _gather_netbox_call,
+            ret_map, uri, fn_name, args,
+            {timeout = opts.timeout}
+        )
+        fiber:name('netbox_call_map')
+        fiber:set_joinable(true)
+        fibers[uri] = fiber
+    end
+
+    for _, uri in pairs(opts.uri_list) do
```
sharonovd: I'd suggest to add ell errors as second example - for simple cases when it invalidates the whole request. Otherwise user will always have to grind through result map searching for errors

rosik: Is your proposal to return something like this?

```lua
  {3301 = ret3301, 3302 = nil},
  {3301 = nil, 3302 = err3302},
```

sharonovd: Yes, something like this. I just want a quick way to understand if all was well

rosik:
Merged in #332.

PR #248, opened 2019-10-04
Fix graphql.validate.getParentFields

HustonMmmavr:
Now it doesn't fail and and returns nil, if there are no fields in object

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Leader flag in server details modal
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -38,6 +38,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Total buckets count in buckets tooltip
 
+- Graphql `validate.getParentField()` doesn't fail and returns nil, if there 
+  are no fields in parent object
+
```
rosik:
```suggestion
- Human-readable error for invalid GrqphQL queries:  `Field "x" is not defined on type "String"`
```

PR #255, opened 2019-10-07
Make edit_topology graphql args optional

rosik:
Uuid is optional when mutation is used to create new replicaset


rosik:
Cherry-picked in #232

PR #258, opened 2019-10-08
Ready tests:

Errora:
- Create replicasets and Bootstrap,
- Join Replica Set,
- Edit replica set,
- ProbeServer,
- Failover,
- Expel server.

Start: Detail server, Login, Auth


webui/src/components/AuthToggleButton.js
```diff
@@ -36,8 +36,7 @@ class AuthToggleButton extends React.Component<AuthToggleButtonProps, {visible:
   render() {
     const {
       implements_check_password,
-      authorizationEnabled,
-      className
```
rgbabaev: It's unwanted change, don't do it

PR #292, opened 2019-10-23
API for setting and querying DDL 

### LINKED ISSUE #312: API for setting and querying DDL

HustonMmmavr:
DDL [schema](https://github.com/tarantool/ddl/blob/dev/README.md) as a rendered YAML string is stored as a section in clusterwide config (with the section name `schema.yml`). Being a part of clusterwide config, it's updated with two-phase commit.

Schema managements doesn't work before the cluster was bootstrapped.

### DDL manager

This patch introduces new module - ddl manager. It's used for config validation and applying in the manner other roles do.

Validation stage calls `ddl.check_schema()` (which checks space existence and checks creation of dummy space). Additionally, validation stage should check that no spaces were removed (in the future it'll be the responsibility of migrations). 

Apply phase calls `ddl.set_schema()` and creates spaces. To avoid replication conflicts schema validation and application are run on leaders only.

### GraphQL API

Cartridge provides GraphQL API for managing the schema:

```graphql
query get_schema {
  cluster {
    schema {
      as_yaml # get yaml representation of the schema
      # accessing spaces one by one may be implemented later, but not now
    }
  }
}
```

```graphql
mutation set_schema($yaml: String!) {
  cluster {
    schema(as_yaml: $yaml) {
      as_yaml # returns new schema
      # return type of mutation must the same graphql type as of get_schema query
    }
  }
}
```

```graphql
mutation check_schema($yaml: String!) {
  cluster {
    check_schema(as_yaml: $yaml) {
      error # null if OK, message string otherwise.
      # other fields e.g. line and column may be added later
      # if we find the way to calculate it
    }
  }
}
```

To support checking schema on read-only nodes it's executed remotely using `netbox` `eval`.

### HTTP API

Cartridge doesn't provide dedicated HTTP methods for the schema management, but user can access it through the clusterwide config - HTTP `/admin/config` [GET](https://github.com/tarantool/cartridge/blob/df444c93d533a82387a999080c83838648de4810/cartridge/webui/api-config.lua#L40) and [PUT](https://github.com/tarantool/cartridge/blob/df444c93d533a82387a999080c83838648de4810/cartridge/webui/api-config.lua#L69) methods.

### Lua API

Cartridge exports two ***global*** Lua functions:

* `_G.cartridge_get_schema()` - retrieves schema from the clusterwide config (the same yaml string)
* `_G.cartridge_set_schema(schema_yml)` - set schema, triggering `patch_clusterwide`. Returns the same schema set.

### Example clusterwide config

```yaml
schema: |
  ---
  spaces:
    A: {"format": ..., "indexes": ...}
    B: ...
  ...
```

Schema specification is described [here](https://github.com/tarantool/ddl/blob/dev/README.md)

Close #312

### DIFF
--- a/cartridge/confapplier.lua
+++ b/cartridge/confapplier.lua
@@ -16,6 +16,7 @@ local pool = require('cartridge.pool')
 local utils = require('cartridge.utils')
 local roles = require('cartridge.roles')
 local topology = require('cartridge.topology')
+local ddl_manager = require('cartridge.ddl-manager')
 local service_registry = require('cartridge.service-registry')
 
 yaml.cfg({
@@ -212,6 +213,11 @@ local function validate_config(conf_new, conf_old)
     end
     checks('table', 'table')
 
+    local ok, err = ddl_manager.validate_config(conf_new, conf_old)
+    if not ok then
+        return nil, err
+    end
+
     return roles.validate_config(conf_new, conf_old)
 end
 
@@ -317,6 +323,12 @@ local function apply_config(conf)
     })
     if err then
         log.error('Box.cfg failed: %s', err)
+    end
+
+    local _, _err = ddl_manager.apply_config(conf, {is_master = is_master})
+    if _err then
+        log.error('%s', _err)
+        err = err or _err
     end
 
     local _, _err = roles.apply_config(conf, {is_master = is_master})
--- a/cartridge/ddl-manager.lua
+++ b/cartridge/ddl-manager.lua
@@ -1,0 +1,133 @@
+#!/usr/bin/env tarantool
+
+local log = require('log')
+local ddl = require('ddl')
+local yaml = require('yaml')
+local checks = require('checks')
+local errors = require('errors')
+
+local topology = require('cartridge.topology')
+
+local CheckSchemaError = errors.new_class('CheckSchemaError')
+
+local _section_name = 'schema.yml'
+
+local function _from_yaml(schema_yml)
+    if schema_yml == nil then
+        return {}
+    elseif type(schema_yml) ~= 'string' then
+        local err = CheckSchemaError:new(
+            'Section %q must be a string, got %s',
+            _section_name, type(schema_yml)
+        )
+        return nil, err
+    end
+
+    local ok, schema_tbl = pcall(yaml.decode, schema_yml)
+    if not ok then
+        local err = CheckSchemaError:new(
+            'Invalid YAML: %s',
+            schema_tbl
+        )
+        return nil, err
+    end
+
+    if type(schema_tbl) ~= 'table' then
+        local err = CheckSchemaError:new(
+            'Schema must be a table, got %s',
+            type(schema_tbl)
+        )
+        return nil, err
+    end
+    return schema_tbl
+end
+
+local function apply_config(conf, opts)
+    checks('table', {is_master = 'boolean'})
+
+    if not opts.is_master then
+        return
+    end
+
+    if conf[_section_name] == nil then
+        return
+    end
+
+    local schema, err = _from_yaml(conf[_section_name])
+    if schema == nil then
+        return nil, err
+    end
+
+    local ok, err = ddl.set_schema(schema)
+    if not ok then
+        return nil, err
+    end
+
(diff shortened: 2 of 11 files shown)


cartridge/ddl.lua
```diff
@@ -0,0 +1,68 @@
...
+
+    if type(new_schema.spaces) ~= 'table' then
+        return nil, e_set_schema:new('Schema spaces is not a table')
+    end
+
+    for space_name, _ in pairs(old_schema.spaces) do
+        if new_schema.spaces[space_name] == nil then
+            return nil, e_set_schema:new(
+                string.format(
+                    "DLD can't remove spaces, but schema doesn't contains space %q",
```
rosik: typo

.gitlab-ci.yml
```diff
@@ -1,3 +1,6 @@
+variables:
+  GIT_SUBMODULE_STRATEGY: recursive
+
```
rosik: redundant

.gitlab-ci.yml
```diff
@@ -31,6 +34,7 @@ build:
     - tarantoolctl rocks install luacheck
     - tarantoolctl rocks install luacov
     - tarantoolctl rocks install luacov-console 1.1.0
+    - tarantoolctl rocks make --chdir third_party/ddl/
```
rosik: redundant

cartridge/ddl.lua
```diff
@@ -0,0 +1,68 @@
...
+    local ddl_schema = confapplier.get_readonly('schema')
+    if ddl_schema == nil then
+        if confapplier.get_readonly() == nil then
+            return nil, e_get_schema:new("Cluster isn't bootstaraped yet")
+        end
+    end
+    return ddl_schema or {spaces = {}}
+end
+
+local function set_schema(new_schema)
```
rosik: Use checks

cartridge/confapplier.lua
```diff
@@ -428,6 +431,19 @@ local function validate_config(conf_new, conf_old)
         end
     end
 
+    if conf_new.schema ~= nil then
```
rosik: conf_new.schema should be stored as a string. Do yaml.decode before validation

cartridge/ddl-manager.lua
```diff
@@ -0,0 +1,133 @@
...
+        return {}
+    elseif type(schema_yml) ~= 'string' then
+        local err = CheckSchemaError:new(
+            'Section %q must be a string, got %s',
+            _section_name, type(schema_yml)
+        )
+        return nil, err
+    end
+
+    local ok, schema_tbl = pcall(yaml.decode, schema_yml)
```
knazarov: use

`CheckSchemaError:pcall()`

cartridge/twophase.lua
```diff
@@ -293,6 +294,33 @@ local function patch_clusterwide(patch)
...
+    local schema_yml = confapplier.get_readonly('schema.yml')
+
+    if schema_yml == nil then
+        return {spaces = {}}
+    else
+        return yaml.decode(schema_yml)
+    end
+end
+
+function _G.cartridge_set_schema(schema)
```
knazarov: this should take YAML

cartridge/twophase.lua
```diff
@@ -293,6 +294,33 @@ local function patch_clusterwide(patch)
     return ok, err
 end
 
+function _G.cartridge_get_schema()
```
knazarov: this should return YAML

test/integration/ddl_test.lua
```diff
@@ -0,0 +1,224 @@
...
+--------------------------------------------------------------------------------
+
+function g.test_luaapi()
+    local function call(...)
+        return g.cluster.main_server.net_box:call(...)
+    end
+
+    log.info('Applying valid schema...')
+
+    t.assert_equals(
```
knazarov: You should test here that you get the same YAML as you push to the API

cartridge/twophase.lua
rosik: The intention was to provide API useful for connectors. They don't care about formatting.

PR #293, opened 2019-10-23
Refactor instance bootstrapping without polling config

rosik:
### Further improvements

- [x] Test cluster cookie chenge after restart
- [x] Make use of confapplier state in 2pc prepare/commit
- [ ] Log patch that failed
- [ ] Test starting two instances in the same workdir
- [x] Test case: boot_instance fails due to invalid memtx memory
- [x] Test case: call edit_topology joining two servers with repeating url
- [x] Test case: check all_rw works with failover enabled


filonenko-mikhail:
clusterwide-config.lua in rockspec missed

rosik:
Closing in favor of #331

PR #305, opened 2019-10-24
#301 add information aboutn clocks difference between cluster members

### LINKED ISSUE #301: Add information into API about cluster members clocks synchronization
Since membership 2.2.0, the members table contains information about clocks difference between each server and other servers in the cluster.
In order to use this information in applications to display alerts if clocks varying exceed some limits, cartridge should add some metrics (at this point, min/max clock delta is enough) into API (lua / graphql)

avkapustin:
Closes #301

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ coverage_result.txt
 .DS_Store
 .vscode
 luacov.stats.out
+build/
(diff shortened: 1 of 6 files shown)


rosik:
Resolved slightly the other way in #409

PR #321, opened 2019-10-30
Add validate tests

Errora:


webui/cypress/integration/validate-tests.spec.js
```diff
@@ -0,0 +1,82 @@
...
+  })
+
+  it('2. Probe server dialog',() => {
+    cy.get('.meta-test__ProbeServerBtn').click();
+    cy.get('.ProbeServerModal input[name="uri"]')
+      .type(' ');
+    cy.get('.meta-test__ProbeServerSubmitBtn').click();//component:ProbeServerModal
+    cy.get('.ProbeServerModal').contains('Probe " " failed: parse error');
+    cy.get('.ProbeServerModal input[name="uri"]')
+      .type('local');
```
rosik: Let"s extend test cases
* press escape, check that dialog is closed
* Ping localhost:9 (or some other unreacable host), Check that error message says "no response" (dialog should still be open)
* Enter uri already joined, press enter, check that the dialog is closed and that success is notified

PR #334, opened 2019-11-14
Move Auth switcher to cluster main page

### LINKED ISSUE #330: "Enable auth" button placement
Cartridge supports custom auth backends. It provides a default one (which stores users in clusterwide config), but in some cases it may be implemented differently (via ldap or disabled at all).

When user management is disabled WebUI looks creepy and useless:

image

Cartridge indicates current user management functionality in GraphQL API
https://github.com/tarantool/cartridge/blob/437f9e8cfcb2bc422c21759756802459c1f94806/doc/schema.graphql#L364-L384

Initially we were going to hide user management page when `implements_list_users == false`, but I can't find a version with this behavior. Let's finally fix it.

See also #1, #182

rgbabaev:
Now we don't show Users page when
```
implements_add_user = false
implements_check_password = true
implements_list_users = false
```
and show auth switcher in cluster main page

Close #330

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   role available" anymore, empty table is returned instead.
   **(incompatible change)**
 
+- Don't show Users page and move auth switcher to cluster page when only `implements_check_password = true`
+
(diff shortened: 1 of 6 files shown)


KlonD90:
Auth button should be only in one place. So it should be on the main page in every case.

Issue #306, opened 2019-10-25
Add all_rw replicaset checkbox in webui


rosik:
Since the first public release v1.0.0 cartridge implements `replicaset.all_rw` parameter, which makes all instances in the replicaset writable. It would be nice to see this option configurable and displayed in the WebUI.

We need:

- indicator on the cluster page
- checkbox in the "Edit replicaset" dialog

Things to be done:
- [ ] design
- [ ] frontend implementation
- [ ] cypress tests

rvrsrec:
Added checkbox "all_rw" in the "Add / Edit replicaset" dialog
![TT_C - All_RW (modal)](https://user-images.githubusercontent.com/22233143/69868052-ba2d4480-12b9-11ea-85d6-c62dd111f2de.png)

rosik:
To be honest, it looks messy

rvrsrec:
Based on the results of joint discussions on Dec 2
![TT_C - All_RW (modal)](https://user-images.githubusercontent.com/22233143/70065643-807f7500-15fc-11ea-852c-351941a4e85c.png)

rosik:
Resolved in #366

PR #366, opened 2019-11-26
all_rw checkbox

### LINKED ISSUE #288: "Vshard Group" and "Replica Set Weight" when creating a replica set
### Create replica set

1. Current modal window:
create replica set 1

Problems:
- the window has a large number of roles;
- blocks with weight and group are not visible;
- the group is not selected;
- the button "create replica set" is blocked, the user does not understand what is wrong

Decision:
- move the weight and group blocks to the top of the window

2. Future modal window:
create replica set 2

"Vshard Group" and "Replica Set Weight" in all screens of modal windows moves up

rgbabaev:
- Group and weight fields moved bottom in replicaset modals.
- Added all_rw field.
- Updates UI-Kit to 0.6.0
- Enhance modals scrolling

I did'n forget about

- [x] Tests
- [x] Changelog
- [x] Documentation (Not needed)

Close #288, #306, #380 

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Instances now have internal state machine which helps to manage
   cluster operation and protect from invalid state transitions.
 
+- WebUI checkbox to specify `all_rw` replicaset property.
+
 ### Changed
 
 - WebUI now uses `edit_topology` mutation instead of deprecated ones.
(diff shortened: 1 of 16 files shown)


rosik:
Don't forget to provide screenshots, let people see them

image

"Edit replicaset" dialog is OK.

Indicator in replicaset list is needed too. It could be ether the same "all_rw" indicator or readonly indication on every server in list. Let's ask @rvrsrec what he thinks is better.

Don't close #306 yet unless you implement tests.

P.S. Pay attention to server details dialog: general -> ro. It's always empty (it's likely a bug)

rosik:
And this is how new scrolling looks like

image

PR #375, opened 2019-11-28
rpc: support on_push and on_push_ctx options for net.box call

olegrok:
See https://www.tarantool.io/ru/doc/1.10/book/box/box_session/#box-session-push
for details


rosik:
Does `on_push` work for `netbox.self`?

olegrok:
> Does `on_push` work for `netbox.self`?

https://github.com/tarantool/tarantool/issues/4657

No. And I suppose that it will be never fixed.

Issue #95, opened 2019-08-28
Split graphql into a separate rock


sharonovd:
In GitLab by @rosik on May 6, 2019, 18:08

olegrok:
Seems that we should move our graphql to tarantool/graphql repository.

For several reasons:
- cartridge/graphql contains improved variable validation cherry-picked from https://github.com/tarantool/graphql.0
- cartridge/graphql is actively supported in contrast of https://github.com/tarantool/graphql

PR #376, opened 2019-11-28
Enable luacheck for graphql files

RunsFor:
Also fixed existed warnings

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -8,6 +8,4 @@ include_files = {
 }
 exclude_files = {
     '.rocks',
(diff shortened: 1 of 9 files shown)


rosik:
Graphql code doesn't belong to this repo. It was copy-pasted long time ago from TDG and from https://github.com/tarantool/graphql/. If you want to enhance it, do it in that repo.

Also note #95

rosik:
Recently in #745 we enabled graphql testing on cartridge side. I think it's appropriate now to fix linter as well.

PR #424, opened 2019-12-20
Retry bootstrapping if snapshot is missing

rosik:
According to the users feedback, the common way to repair the
replication is to remove snapshots and rebootstrap the replica. This
behavior was changed when refactoring bootstrap process, but this patch
brings it back.

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)

### DIFF
--- a/test/integration/reboot_test.lua
+++ b/test/integration/reboot_test.lua
@@ -90,22 +90,47 @@ function g.test_absent_config()
     )
     log.warn('Config removed')
 
-    g.cluster.main_server:start()
-    g.cluster:retrying({}, function()
-        g.cluster.main_server:connect_net_box()
-    end)
+    local srv = g.cluster.main_server
+    srv:start()
 
-    local state, err = g.cluster.main_server.net_box:eval([[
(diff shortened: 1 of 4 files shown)


test/integration/reboot_test.lua
```diff
@@ -90,31 +90,32 @@ function g.test_absent_config()
...
+    end)
 end
 
 function g.test_absent_snapshot()
     g.cluster:stop()
     log.warn('Cluster stopped')
 
     local workdir = g.cluster.main_server.workdir
-    for _, f in pairs(fio.glob(fio.pathjoin(workdir, '*.snap'))) do
+    for _, f in pairs(fio.glob(workdir .. '/*.snap')) do
```
filonenko-mikhail: beauty degradation(

rosik: Fixed

rosik:
@knazarov requested a change: We should be able to remove leaders snapshot, and if the replica is already configured, the roles should be switched (leader should be joined to the replica)

PR #433, opened 2019-12-24
Avoid using netbox.self in RPC

### LINKED ISSUE #422: Avoid netbox.self in rpc_call({prefer_local = false})
Using `netbox.self` is full of problems: see #377, #208, https://github.com/tarantool/tarantool/issues/4657, https://github.com/tarantool/tarantool/issues/4602, https://github.com/tarantool/tarantool/issues/4540, https://github.com/tarantool/tarantool/issues/4513.

We can't stop using it because it's essential for `prefer_local=true`, but I think it would be fair to avoid it when user specifies `prefer_local = false` explicitly.

DHorrible:
Avoid use `netbox.self` for `{prefer_local=false}` option on `rpc_call`

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #422

### DIFF
--- a/test/integration/rpc_test.lua
+++ b/test/integration/rpc_test.lua
@@ -119,6 +119,13 @@ function g.test_routing()
     t.assert_equals(res.uuid, B2.instance_uuid)
     t.assert_equals(res.peer, B2.net_box:call('box.session.peer'))
 
+    local res, err = rpc_call(B2,
+        'myrole', 'get_session', nil,
+        {prefer_local = false}
+    )
+    t.assert_not(err)
+    t.assert_not_equals(res.peer, B2.net_box:call('box.session.peer'))
+
     -- Test opts.leader_only
     --------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


test/integration/rpc_test.lua
```diff
@@ -119,6 +119,13 @@ function g.test_routing()
     t.assert_equals(res.uuid, B2.instance_uuid)
     t.assert_equals(res.peer, B2.net_box:call('box.session.peer'))
 
+    local res, err = rpc_call(B2,
+        'myrole', 'get_session', nil,
+        {prefer_local = false}
+    )
+    t.assert_not(err)
+    t.assert_not_equals(res.peer, B2.net_box:call('box.session.peer'))
```
rosik: The test doesn't always reproduce an error even without a fix. The `rpc_call` may be performed on B1, and, of course, peer would differ in this case.

PR #466, opened 2020-01-14
Fix 2pc backward compatibility with 1.2.0

rosik:
To update cartridge without downtime it's necessary to perform rolling
upgrade: restart half a cluster with new version, switch leaders, then
restart the other half.

In #331 we've changed naming of two-phase commit internal functions,
so compatibility between old and new version was broken and it was
impossible to switch the leadership during upgrade.

This patch brings the compatibility back and covers rolling upgrade
scenario with luatest.

I didn't forget about

- [x] Tests
- [x] Changelog

### DIFF
--- a/test/compatibility/upgrade_test.lua
+++ b/test/compatibility/upgrade_test.lua
@@ -1,0 +1,209 @@
+local fio = require('fio')
+local log = require('log')
+local yaml = require('yaml')
+local fiber = require('fiber')
+local errors = require('errors')
+local t = require('luatest')
+local g = t.group()
+
+local test_helper = require('test.helper')
+local helpers = require('cartridge.test-helpers')
+
+g.before_all = function()
+    g.tempdir = fio.tempdir()
+end
+
+g.after_all = function()
+    if g.cluster ~= nil then
+        g.cluster:stop()
+    end
+    if g.tempdir ~= nil then
+        fio.rmtree(g.tempdir)
+        g.tempdir = nil
+    end
+end
+
+local function version(srv)
+    local version = srv.net_box:call('require', {'cartridge.VERSION'})
+    local cwd = srv.net_box:call('package.loaded.fio.cwd')
+    log.info('Check %s version: %s (from %s)', srv.alias, version, cwd)
+    return version
+end
+
+local function upstream_info(srv)
+    local info = srv.net_box:call('box.info')
+    for _, v in pairs(info.replication) do
+        if v.uuid ~= srv.instance_uuid then
+            return {
+                status = v.upstream.status,
+                message = v.upstream.message,
+            }
+        end
+    end
+end
+
+local function wish_state(srv, desired_state)
+    g.cluster:retrying({}, function()
+        srv.net_box:eval([[
+            local confapplier = require('cartridge.confapplier')
+            local desired_state = ...
+            local state = confapplier.wish_state(desired_state)
+            assert(
+                state == desired_state,
+                string.format('Inappropriate state %q ~= desired %q',
+                state, desired_state)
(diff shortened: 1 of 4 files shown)


test/compatibility/upgrade_test.lua
```diff
@@ -0,0 +1,209 @@
...
+        if v.uuid ~= srv.instance_uuid then
+            return {
+                status = v.upstream.status,
+                message = v.upstream.message,
+            }
+        end
+    end
+end
+
+local function wish_state(srv, desired_state)
```
sharonovd: await_state, probably?

test/compatibility/upgrade_test.lua
```diff
@@ -0,0 +1,209 @@
...
+        ]], {desired_state})
+    end)
+end
+
+--- Test upgrading form old version to the current one
+function g.test_upgrade()
+    local cartridge_elder_path = os.getenv('CARTRIDGE_ELDER_PATH')
+    t.skip_if(
+        cartridge_elder_path == nil,
+        'No elder version provided. Skipping'
```
sharonovd: older, probably

test/compatibility/upgrade_test.lua
```diff
@@ -0,0 +1,209 @@
...
+    replica:start()
+    g.cluster:retrying({}, function() replica:connect_net_box() end)
+    wish_state(replica, 'RolesConfigured')
+
+    t.assert_not_equals(version(leader), 'scm-1')
+    t.assert_equals(version(replica), 'scm-1')
+    t.assert_equals(upstream_info(replica), {status = 'follow'})
+
+    -- Switch the leadership
+    leader.net_box:eval([[
```
sharonovd: HTTP API call looks preferrable (more honest)

test/compatibility/upgrade_test.lua
```diff
@@ -0,0 +1,209 @@
...
+        errors.assert('E', cartridge.admin_edit_topology(...))
+    ]], {{
+        replicasets = {{
+            uuid = leader.replicaset_uuid,
+            failover_priority = {replica.instance_uuid},
+        }}
+    }})
+    g.cluster.main_server = replica
+
+    fiber.sleep(0.2)
```
sharonovd: why?

test/compatibility/upgrade_test.lua
```diff
@@ -0,0 +1,209 @@
...
+                string.format('Inappropriate state %q ~= desired %q',
+                state, desired_state)
+            )
+        ]], {desired_state})
+    end)
+end
+
+--- Test upgrading form old version to the current one
+function g.test_upgrade()
+    local cartridge_elder_path = os.getenv('CARTRIDGE_ELDER_PATH')
```
sharonovd: I think this should be on in CI

test/compatibility/upgrade_test.lua
rosik: Agree

test/compatibility/upgrade_test.lua
rosik: Agree

test/compatibility/upgrade_test.lua
rosik: Agree

test/compatibility/upgrade_test.lua
rosik: I've changed it with more clear `g.cluster.retying`. The idea is to wait for workload switching before killing leader

test/compatibility/upgrade_test.lua
rosik: Done

PR #492, opened 2020-01-21
Draft implementation of try-cartridge example

### LINKED ISSUE #454: Design extension role
## The task
We need to build the special extension role which will let users, write code and match user-functions like an HTTP server endpoint or just match function as a stored procedure. This role will let users simply write some functionality (maybe, distributed) and pass it externally.

rosik:
This is an example configuration for try-cartridge demo deployment

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Close #454


test/integration/extensions/banking_config/extensions/banking.lua
```diff
@@ -0,0 +1,30 @@
...
+local function account_add(customer_id, account_id, name)
+    checks('number', 'number', 'string')
+    return box.space.account:insert(
+        {customer_id, account_id, name, 0}
+    )
+end
+
+local function transfer_money(account_from, account_to, amount)
+    box.begin()
+        box.space.account:update({account_to}, {{'+', 4, amount}})
```
artur-barsegyan: Please, return back the normal indentation because I think that the current one is misleading.
You can leave blank lines between the start of the transaction and the end.

test/integration/extensions/banking_config/extensions/banking.lua
```diff
@@ -0,0 +1,30 @@
...
+        box.space.account:update({account_from}, {{'-', 4, amount}})
```
artur-barsegyan: What about the number '4'? Do we have another API without field number?

test/integration/extensions/banking_config/extensions/config.yml
```diff
@@ -0,0 +1,29 @@
+---
+service: banking
+
+provider:
```
artur-barsegyan: What's the purpose to create `service` and `provider` field?

rosik:
Moved to https://github.com/tarantool/cartridge-extensions/

PR #498, opened 2020-01-23
Tweak PHP example in DemoInfo.js

rybakit:
The PHP example is adjusted to follow the [PSR-12](https://www.php-fig.org/psr/psr-12/) coding style recommendations. Besides, `include_once()` with a relative path is replaced with `require __DIR__` to match composer's [documentation](https://getcomposer.org/doc/01-basic-usage.md#autoloading). Also, the `-f` flag in the CLI command is not necessary, so it's removed.

### DIFF
--- a/webui/src/components/DemoInfo.js
+++ b/webui/src/components/DemoInfo.js
@@ -72,9 +72,11 @@ composer require tarantool/client
 
 \`\`\`php
 <?php
-include_once('vendor/autoload.php');
+
 use Tarantool\\Client\\Client;
 use Tarantool\\Client\\Schema\\Criteria;
+
+require __DIR__ . '/vendor/autoload.php';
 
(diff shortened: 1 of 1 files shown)


rybakit:
Looking at the example, I'm not sure if it's worth showing from the very beginning how to select *all* tuples from space (even if there's only one in this example). People may start blindly copying this code and applying it to large spaces, which should certainly be avoided (I deliberately designed the client API to make such things harder to write).

It would be better to show something like this:

```php
$space->insert([222, 'Michael Bryan']);
$result = $space->select(Criteria::key([222]));
```

WDYT?

rybakit:
Also, it seems like the scheme prefix is missing in the DNS string:

```diff
- $client = Client::fromDsn(':demo_uri:');
+ $client = Client::fromDsn('tcp://:demo_uri:');
```

PR #526, opened 2020-01-30
Added http adapter

### LINKED ISSUE #453: Update HTTP server to v2.0

Satbek:
* http updated to v2.1.0 
* Added http adapter. It is a layer which provides http-v1 api to interact with http-v2

Closes #453


Satbek:
The http versions are not compatible, so http-adapter fixes it. It's possible to interact with ``http v2`` as with ``http v1`` via ``cartridge.http-adapter``.  

``http-adapter`` is available in ``cartridge.service_get('httpd')``
``http v2``  is available in ``cartridge.service_get('httpd-2')``. 

``cartridge.service_get('httpd-2')`` returns ``http.server`` object, which has ``http.router`` with default options.

The differences between **http v1** and **http v2** APIs are:
- splitted router and server objects in ``http v2``, in ``http v1`` they are parts of one object
- before_dispatch and after_dispatch hooks are not supported by default
- headers, path, peer, method, proto, query in ``http v1`` are request's object fields, but they are methods in ``http v2`` 

So based on this:
- http-adapter contains http.server and http.router objects and redirects requests(methods calls, fields updates) to them
- http-adapter decorates all route's handlers and adds hooks logic
- http-adapter replaces every request object in route handler's arguments


``http-adapter`` passes ``http v1.1.0`` tests.

rosik:
We decided to postpone it to the future

PR #546, opened 2020-02-04
Remember last opened file in local storage

### LINKED ISSUE #504: [TC] Open the first file from the file tree at the "Code" page
After that, it shows only the last opened.
Save it only per current session - after page reboot, it needs to be flushed

rgbabaev:
* choose file to open at first run

* file tree expanded by default

I didn't forget about

- [x] Tests
- [x] Changelog
- [ ] Documentation

Close #504

### DIFF
--- a/webui/src/components/FileTree/index.js
+++ b/webui/src/components/FileTree/index.js
@@ -27,6 +27,7 @@ type FileTreeProps = {
   operationObject?: ?string,
   tree: Array<TreeFileItem>,
   selectedFile: FileItem | null,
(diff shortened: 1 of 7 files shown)


webui/src/components/FileTree/index.js
```diff
@@ -27,6 +27,7 @@ type FileTreeProps = {
   operationObject?: ?string,
   tree: Array<TreeFileItem>,
   selectedFile: FileItem | null,
+  initiallyExpaned: boolean,
```
KlonD90: Expanded?

Issue #542, opened 2020-02-03
Describe cartridge error handling gudelines


rosik:
How are errors usually reported (in lua/graphql/http/internally)?
How should I check errors returned from cartridge public API?

Short answer:

Cartridge uses https://github.com/tarantool/errors module. Cartridge doesn't throw exceptions (except for two cases: contract violation in function calls, and bugs).

GraphQL errors: see #491 and https://spec.graphql.org/June2018/#sec-Errors

HTTP: [RTFS](https://github.com/tarantool/cartridge/blob/master/cartridge/webui/api-config.lua#L60)

rosik:
Also it would be nice to mention different APIs: GraphQL, Lua, HTTP.

Debt: describe GraphQL error extensions and how they can be specified in error objects.

olegrok:
> Debt: describe GraphQL error extensions and how they can be specified in error objects.

Example: https://github.com/tarantool/cartridge/pull/575/files#diff-cf7848a656b50a8b4ce2cc926bef5205R205

rosik:
Also related: https://github.com/tarantool/errors

PR #575, opened 2020-02-13
Implement GraphQL on_resolve triggers

### LINKED ISSUE #568: Implement callback to cancel graphql execution if needed
#### My proposal

User code:

```
local function allow_execution(is_mutation, prefix, name, req)
    ...
    return true
end
```

Cartridge:

```
local function set_execution_callback(callback)
    ...
end

local function execute_graphql(req)
    ...
    if execution_permission_callback ~= nil then
        local _, err = execution_permission_callback(...)
        if err ~= nil then
            ...
        end
    ...
end

return {
    ...
    set_execution_callback = set_execution_callback,
    ...
}
```

rosik:
It will be executed **before** top-level resolvers, which were
registered using `add_callback` or `add_mutation` methods.

The trigger function is called with two arguments:
- `operation` (*string*): `"query|mutation"`,
- `field` (*string*): `"[prefix.]field_name"`.

If the parameters are `(nil, old_trigger)`, then the old trigger is deleted.

Example:

```lua
local function log_request(operation, field)
    log.info('GraphQL %s: %s', operation:upper(), field)
    -- Will print "GraphQL QUERY cluster.auth_params"
end)

graphql.on_resolve(log_request) -- start logging
graphql.on_resolve(nil, log_request) -- stop logging
```

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)
- [x] Documentation

Close #568

### DIFF
--- a/cartridge/graphql.lua
+++ b/cartridge/graphql.lua
@@ -35,14 +35,11 @@ local function get_fields()
     return vars.graphql_schema_fields
 end
 
-local function funcall_wrap(fun_name, trigger_arg)
-    checks('string', {
-        operation = 'string',
-        field = 'string',
-    })
+local function funcall_wrap(fun_name, operation, field_name)
+    checks('string', 'string', 'string')
     return function(...)
         for trigger, _ in pairs(vars.on_resolve_triggers) do
-            trigger(trigger_arg)
+            trigger(operation, field_name)
         end
 
         local res, err = funcall.call(fun_name, ...)
@@ -117,10 +114,9 @@ local function add_callback(opts)
         oldkind.fields[opts.name] = {
             kind = opts.kind,
             arguments = opts.args,
-            resolve = funcall_wrap(opts.callback, {
-                operation = 'query',
-                field = opts.prefix .. '.' .. opts.name,
-            }),
+            resolve = funcall_wrap(opts.callback,
+                'query', opts.prefix .. '.' .. opts.name
+            ),
             description = opts.doc,
         }
 
@@ -133,10 +129,9 @@ local function add_callback(opts)
         vars.callbacks[opts.name] = {
             kind = opts.kind,
             arguments = opts.args,
-            resolve = funcall_wrap(opts.callback, {
-                operation = 'query',
-                field = opts.name,
-            }),
+            resolve = funcall_wrap(opts.callback,
+                'query', opts.name
+            ),
             description = opts.doc,
         }
     end
@@ -162,10 +157,9 @@ local function add_mutation(opts)
         oldkind.fields[opts.name] = {
             kind = opts.kind,
             arguments = opts.args,
-            resolve = funcall_wrap(opts.callback, {
-                operation = 'mutation',
-                field = opts.prefix .. '.' .. opts.name,
-            }),
+            resolve = funcall_wrap(opts.callback,
+                'mutation', opts.prefix .. '.' .. opts.name
+            ),
             description = opts.doc
         }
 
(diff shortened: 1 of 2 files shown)


cartridge/graphql.lua
```diff
@@ -346,6 +367,42 @@ local function init(httpd)
...
+--
+-- @usage
+--    local function log_request(req)
+--        log.info('GraphQL %s: %s', req.operation:upper(), req.field)
+--        -- Will print "GraphQL QUERY cluster.auth_params"
+--    end)
+--
+--    graphql.on_resolve(log_request) -- start logging
+--    graphql.on_resolve(nil, log_request) -- stop logging
+--
```
olegrok: graphql.on_resolve() - should return list of triggers

If we want to respect tarantool triggers flow

rosik: I thought about it, but decided not to overload the patch without necessity. Also I don't like tarantool triggers flow because it's return values are ambiguous.

cartridge/graphql.lua
```diff
@@ -346,6 +367,42 @@ local function init(httpd)
     )
 end
 
+--- Set up trigger for GraphQL handlers.
+--
+-- It will be executed **before** top-level resolvers, which were
+-- registered using `add_callback` or `add_mutation` methods.
+--
+-- The trigger function is called with one argument:
```
olegrok: Why only one?

I suggest you pass it as separate arguments (without any temporary tables) and pass http request object as well

olegrok:
You've added a good description to your PR and only two sentences into commit message

I think it's a good feature when you can use commit messages as documentation

And I don't know about conventions but you could use TarantoolDocBot for doc requests

test/integration/graphql_test.lua
```diff
@@ -192,3 +192,67 @@ function g.test_error_extensions()
...
+function g.test_middleware()
+    local server = cluster.main_server
+
+    -- GraphQL execution can be interrupted by raising error in trigger
+    server.net_box:eval([[
+        local errors = require('errors')
+        graphql = require('cartridge.graphql')
+        function raise()
+            local err = errors.new('E', 'You are not welcome here')
+            err.graphql_extensions = {code = 403}
```
olegrok: This feature is described nowhere

cartridge/graphql.lua
```diff
@@ -34,8 +35,13 @@ local function get_fields()
     return vars.graphql_schema_fields
 end
 
-local function funcall_wrap(fun_name)
+local function funcall_wrap(fun_name, operation, field_name)
+    checks('string', 'string', 'string')
     return function(...)
+        for trigger, _ in pairs(vars.on_resolve_triggers) do
```
mtrempoltsev: I think using an array is redundant

rosik: This interface is native for Tarantool. I could agree with you in this particular case, but I don't want to invent new trigger style.

test/integration/graphql_test.lua
rosik: There is no place to describe it yet, it's a debt. I've mentioned it in #542

cartridge/graphql.lua
olegrok: Hm, no. It's a bit not native.

- In your case setting the trigger is idempotent:
```
tarantool> box.space.log:before_replace(x)
---
- 'function: 0x010ce83f78'
...

tarantool> box.space.log:before_replace(x)
---
- 'function: 0x010ce83f78'
...

tarantool> box.space.log:before_replace()
---
- - 'function: 0x010ce83f78'
  - 'function: 0x010ce83f78'
...
```

- You don't give any guarantees about order of the execution

olegrok: But I agree that ability to setup several triggers could be useful

rosik:
> You've added a good description to your PR and only two sentences into commit message
> 
> I think it's a good feature when you can use commit messages as documentation
> 

Commit message and diff is one whole. I don't see a reason to duplicate most of the information. 

> And I don't know about conventions but you could use TarantoolDocBot for doc requests

DocBot is only useful when there is no ldoc. I prefer writing documentation along with the code rather than creating new issues on github.

PR #584, opened 2020-02-17
Minor text edits to connect info

lenkis:
Brushed up phrasing and markup.

TODO: update tests


lenkis:
Tests in `cartridge` repo do not include the text I changed (grep returns nothing). But these changes may affect some GUI tests, that's why @KlonD90 in approvers.

PR #611, opened 2020-02-27
Add option for clusterwide env in test helpers

printercu:


test/unit/test-helpers/cluster_test.lua
```diff
@@ -0,0 +1,31 @@
+local t = require('luatest')
+local g = t.group()
```
rosik: I would prefer it was added to the existing `helpers_test.test_cluster_helper`

cartridge/test-helpers/cluster.lua
```diff
@@ -69,9 +70,21 @@ function Cluster:new(object)
...
+                    server_config.env[k] = v
+                end
+            end
+        end
+    end
+
     self.servers = {}
     for _, replicaset_config in ipairs(self.replicasets) do
         for i, server_config in ipairs(replicaset_config.servers) do
+            merge_server_env(server_config)
```
rosik: Purpose of this function is unclear without reading it's body. 
How about something like this or similar for your taste?
```lua
table_supplement(server_config.env, self.env or {})
```

It is already used in argparse module: https://github.com/tarantool/cartridge/blob/47db09d51a0459b375a77ec8f98331ec841adc74/cartridge/argparse.lua#L294-L303

(We could move it to utils and reuse)

printercu: Let me replace it with `fun.chain(self.env or {}, server_config.env or {}):tomap()`

PR #689, opened 2020-03-26
graphql: strict type validation

olegrok:
This patchset is a port of https://github.com/tarantool/graphql.0/pull/180 by @Totktonada

See commit messages for the references.

This patchset solves a problem then we don't validate arguments that are passed via variables for nullability-non-nullability

### DIFF
--- a/cartridge/graphql/execute.lua
+++ b/cartridge/graphql/execute.lua
@@ -2,20 +2,7 @@ local path = (...):gsub('%.[^%.]+$', '')
 local types = require(path .. '.types')
 local util = require(path .. '.util')
 local introspection = require(path .. '.introspection')
-
-local function typeFromAST(node, schema)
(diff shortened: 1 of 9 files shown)


Totktonada:
AFAIR, query_util was extracted in graphql.0 to reuse it in avro_schema generation for a query. Maybe you don't need to move the code here.

I had several non-standard types: Long, Map, InputUnion, InputMap. Please, verify that they don't leak here.

Please also consider to to open a PR to https://github.com/tarantool/graphql.

olegrok:
@Totktonada 

> Please also consider to to open a PR to https://github.com/tarantool/graphql.

Will do

olegrok:
Related patch to https://github.com/tarantool/graphql - https://github.com/tarantool/graphql/pull/7

PR #696, opened 2020-03-27
adapt tests for strict mode

### LINKED ISSUE #693: test: Fix tests running in debug mode
Tarantool in debug mode enables strict by default. This lead to some unexpected results in some tests. 

E.g. unit.remote_control.test_call:

```
1) unit.remote_control.test_call
...bin/Projects/cartridge/test/unit/remote_control_test.lua:461: Error message does not contain: "Procedure '' is not defined"
Error message received: ".../o.babin/Projects/cartridge/cartridge/remote-control.lua:83: variable '' is not declared"
```

As we see we try to call non-existent procedure, but error that such function doesn't exists. And there are some similar places.

olegrok:
The main problem - some tests failed on debug mode, another tests
gave unpredictable result on OS X.

Closes #693

### DIFF
--- a/test/unit/remote_control_test.lua
+++ b/test/unit/remote_control_test.lua
@@ -1,3 +1,5 @@
+require('strict').off()
+
 local t = require('luatest')
 local g = t.group()
 
--- a/test/unit/remote_control_test.lua
+++ b/test/unit/remote_control_test.lua
@@ -1,5 +1,3 @@
-require('strict').off()
-
 local t = require('luatest')
 local g = t.group()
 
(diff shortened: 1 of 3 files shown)


cartridge/remote-control.lua
```diff
@@ -63,6 +63,30 @@ local function is_callable(fun)
     end
 end
 
+-- rawget implementation that respects metatable methods
```
rosik: The patch doesn't correspond to the PR caption "adapt tests for strict mode". It would be better if remote-control module remained unchanged.

olegrok: But remote_control test suite can't be launched in strict mode without this change.

olegrok: Ok

I've disabled strict mode for remote_control tests. It seems one way to provide equal test messages for all Tarantool builds

test/unit/remote_control_test.lua
```diff
@@ -1,3 +1,5 @@
+require('strict').off()
```
rosik, line 1: Luatest loads all files before running tests. By this line you disable strict mode everywhere in luatest

olegrok: Yes. I know.

Do you propose to do similar to?
```lua
g.test = function()
strict.off()
-- code ...
strict.on()
end
```

PR #698, opened 2020-03-27
Gracefully leave membership on shutdown

### LINKED ISSUE #659: Leave membershit at_exit
Nowadays stopping an instance with Ctrl+C results in "dead" status. In some way it does make sense, but it takes too long.
Membership module has the feature - `leave()`, which broadcasts special status immediately.
And cartridge should use it. It would speed up failover triggering and state propagation.

olegrok:
Setup on_shutdown trigger (available since Tarantool 2.1) and
gracefully leave membership at exit.

Closes #659

- [x] Tests


rosik:
I've revised your patch, check it again please.

1. on_shutdown trigger should never raise. Otherwise it'll hang and can only be killed with sigkill. That was the case in `integration.state_machine.test_leader_death`, which leaves membership in order to simulate quick failure, but due to the bug in membership, `leave` can't be called twice.
2. Separate test group is unnecessary in my opinion, there are already many tests with dead instances.

olegrok:
> I've revised your patch, check it again please.

Ok, I agree

PR #699, opened 2020-03-27
don't rebuild graphql schema for every call

### LINKED ISSUE #567: Cache graphql.get_schema result
Now graphql.get_schema calls on each request, but this is needed only if model or callbacks changed.

See: https://github.com/tarantool/cartridge/blob/master/cartridge/graphql.lua#L169

olegrok:
Cache graphql schema

Closes #567

### DIFF
--- a/cartridge/graphql.lua
+++ b/cartridge/graphql.lua
@@ -15,8 +15,7 @@ local funcall = require('cartridge.graphql.funcall')
 local validate = require('cartridge.graphql.validate')
 
 
-vars:new('graphql_schema_fields', {})
-vars:new('graphql_schema', {})
+vars:new('graphql_schema', nil)
 vars:new('model', {})
 vars:new('on_resolve_triggers', {})
 vars:new('callbacks', {})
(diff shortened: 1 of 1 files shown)


cartridge/graphql.lua
```diff
@@ -178,10 +174,16 @@ local function add_mutation(opts)
...
     end
+    vars.graphql_schema = nil
 end
 
 local function get_schema()
-    local fields = table.copy(get_fields())
+    local existing_scheme = vars.graphql_schema
+    if existing_scheme ~= nil then
+        return existing_scheme
+    end
```
rosik, line 184:
```suggestion
    if vars.graphql_schema ~= nil then
        return vars.graphql_schema
    end
```

cartridge/graphql.lua
```diff
@@ -135,6 +130,7 @@ local function add_callback(opts)
             description = opts.doc,
         }
     end
+    vars.graphql_schema = nil
```
rosik, line 133: How about adding the same to `add_callback_prefix` and `add_mutation_prefix`?

cartridge/graphql.lua
olegrok: Applied

cartridge/graphql.lua
olegrok: May be it could improve error message in case if we try to use some "prefix" and there are no any callbacks/mutations (to be honest it's strange case). Fixed.

PR #705, opened 2020-03-30
Cache connection to kingdom for coordinators

### LINKED ISSUE #681: Inactive coordinators spam new connections
Usually there may be several instances with coordinator role enabled, but only one can acquire lock in external storage. In the mean time, other coordinators (inactive) keep trying, but with every attempt to acquire a lock it creates new connection, which is closed soon. This is no good and should be fixed

olegrok:
Closes #681


olegrok:
Should be suppressed by https://github.com/tarantool/cartridge/pull/720

PR #604, opened 2020-02-25
Stateful failover

rosik:
This patch introduces new failover mode - stateful.

There are 3 main concepts:

**Internal coordinator** is a new role, which makes decisions regarding leadership. Earlier it was a part of every instances' failover module, but now it's split away. There may be only one active coordinator in cluster at a time. Its uniqueness is ensured by external storage which manages the lock and saves appointments.

**External storage** (`kingdom.lua`) is a stand-alone Tarantool instance which provides locking mechanism and keeps decisions made by the coordinator.

**Failover module** (the old one) operates on every instance in cluster and gathers leadership information for others modules. It was refactored too, and now it can be described with 4 functions with clearly separated responsibilities:

- `_get_appointments_*` generates leadership map by itself or polls it from external storage depending on the mode setting (disabled/eventual/stateful).
- `accept_appointments()` just refreshes the cache and tracks if anything changed.
- `failover_loop` (a fiber) repeatedly gets new appointments and accepts them using corresponding functions from the above.
- `cfg` is called from `confapplier.apply_config()` (on restart or on committing new clusterwide configuration). At first it gets appointments synchronously and then starts the failover loop.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #148

PR #709, opened 2020-03-30
Forbid reacquiring kingdom lock if it was stolen

rosik:
Here is the case:

1. Coordinator C1 acquires a lock and freezes (but his session remains active);
2. Lock delay expires and kingdom allows C2 to acquire it again;
3. C2 writes some decisions and releases a lock;
4. C1 comes back and tries to renew the lock;

*The goal*: C1 must be informed on his outage.

I didn't forget about

- [x] Tests

Follow-up for #604

### DIFF
--- a/test/integration/kingdom_test.lua
+++ b/test/integration/kingdom_test.lua
@@ -58,7 +58,7 @@ function g.test_locks()
         false
     )
     t.assert_equals(
-        {c2:call('set_leaders', {{'A', 'a1'}})},
+        {c2:call('set_leaders', {{{'A', 'a1'}}})},
         {box.NULL, 'You are not holding the lock'}
     )
     t.assert_equals(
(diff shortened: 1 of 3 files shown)


test/integration/kingdom_test.lua
```diff
@@ -145,3 +145,49 @@ function g.test_passwd()
...
+
+    local payload = {uuid.str(), 'localhost:9'}
+
+    local c1 = connect(g.kingdom)
+    t.assert_equals(
+        {c1:call('acquire_lock', payload)},
+        {true}
+    )
+    t.assert_equals(
+        -- C1 can renew expired lock if it wasn't stolen yen
```
HustonMmmavr: -- C1 can renew expired lock if it wasn't stolen yet

test/integration/kingdom_test.lua
```diff
@@ -145,3 +145,49 @@ function g.test_passwd()
...
+
+    local c2 = connect(g.kingdom)
+    t.assert_equals(
+        {c2:call('acquire_lock', payload)},
+        {true}
+    )
+    c2:close()
+
+    t.assert_equals(
+        -- C1 can't renew lock after if was stolen by C2
```
HustonMmmavr: it?

PR #720, opened 2020-03-31
Fix coordinator architecture

### LINKED ISSUE #680: Coordinator may override manually promoted leader
The coordinator makes decisions basing on local context:

https://github.com/tarantool/cartridge/blob/249f95571d557579aa041a7c6a1ad4a47387ebde/cartridge/roles/coordinator.lua#L58-L68

But leader appointments are saved directly to external storage and not to local context:

https://github.com/tarantool/cartridge/blob/249f95571d557579aa041a7c6a1ad4a47387ebde/cartridge/roles/coordinator.lua#L285-L311

It means manual appointment may me overridden by the next failover triggering.

### LINKED ISSUE #681: Inactive coordinators spam new connections

rosik:
There are several problems:

    1. While external storage is locked by active coordinator others spam
       new connections (#681).
    2. Coordinator may override manually promoted leader (#680).
    3. Manual promotion and take_control_loop are subkect to several race
       conditions.
    4. Coordinator didn't handle URI change properly.

    This patch introduces new module - state provider client. The client is
    created once for one set of parameters and handles reconnections.

    While reconnecting is opaque in state provider client, it's only
    achieved by explicitly calling get_session method. Within a session no
    reconnections are possible. While using the same session the coordinator
    can be sure that no race conditions result in inconsistency.

    The session state machine corresponds to the one of netbox connection:
    initial -> active -> error/closed.

    The client caches the session and creates new one only if the previous
    has died.

I didn't forget about

- [x] Tests

Close #680 
Close #681

### DIFF
--- a/cartridge/roles/coordinator.lua
+++ b/cartridge/roles/coordinator.lua
@@ -16,7 +16,6 @@ vars:new('membership_notification', membership.subscribe())
 vars:new('connect_fiber', nil)
 vars:new('topology_cfg', nil)
 vars:new('client', nil)
-vars:new('session', nil)
 vars:new('options', {
     RECONNECT_PERIOD = 5,
     IMMUNITY_TIMEOUT = 15,
@@ -134,17 +133,14 @@ end
 
 local function take_control(client)
     checks('stateboard_client')
-    assert(vars.session == nil or not vars.session:is_locked())
 
     local lock_args = {
         confapplier.get_instance_uuid(),
         confapplier.get_advertise_uri()
     }
 
-    local session, err = client:get_session()
-    if session == nil then
-        return nil, err
-    end
+    local session = client:get_session()
+    assert(not session:is_locked())
     assert(session.ctx == nil)
 
     local lock_delay, err = session:get_lock_delay()
@@ -177,9 +173,7 @@ local function take_control(client)
     for replicaset_uuid, leader_uuid in pairs(leaders) do
         ctx.decisions[replicaset_uuid] = pack_decision(leader_uuid)
     end
-
     session.ctx = ctx
-    vars.session = session
 
     log.info('Lock acquired')
     --------------------------------------------------------------------
@@ -193,7 +187,7 @@ local function take_control(client)
             break
         end
 
-        if pcall(fiber.status, control_fiber) == 'dead'
+        if fiber.status(control_fiber) == 'dead'
         or not session:acquire_lock(lock_args)
         then
             break
@@ -230,11 +224,6 @@ local function stop()
     if vars.connect_fiber ~= nil then
         pcall(fiber.cancel, vars.connect_fiber)
         vars.connect_fiber = nil
-    end
-
-    if vars.session ~= nil then
-        vars.session:drop()
-        vars.session = nil
     end
 
     if vars.client ~= nil then
@@ -332,9 +321,15 @@ local function appoint_leaders(leaders)
         end
     end
 
-    local session = vars.session
+    if vars.client == nil then
+        return nil, AppointmentError:new("No state provider configured")
+    end
 
-    if session == nil or not session:is_locked() then
+    local session = vars.client.session
+    if session == nil
+    or session.ctx == nil
+    or not session:is_locked()
+    then
         return nil, AppointmentError:new("No active coordinator session")
     end
 
--- a/cartridge/roles/coordinator.lua
+++ b/cartridge/roles/coordinator.lua
@@ -92,7 +92,7 @@ local function control_loop(session)
     checks('stateboard_session')
     local ctx = assert(session.ctx)
 
-    repeat
+    while true do
         ctx.members = membership.members()
 
         local updates = {}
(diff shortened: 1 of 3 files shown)


cartridge/roles/coordinator.lua
```diff
@@ -2,22 +2,21 @@ local log = require('log')
...
 local AppointmentError = errors.new_class('AppointmentError')
 local CoordinatorError = errors.new_class('CoordinatorError')
-local NetboxConnectError = errors.new_class('NetboxConnectError')
 
 vars:new('membership_notification', membership.subscribe())
 vars:new('connect_fiber', nil)
 vars:new('topology_cfg', nil)
-vars:new('conn', nil)
+vars:new('client', nil)
+vars:new('session', nil)
```
olegrok, line 19: If "client" contains "session". I don't see reasons to store session in vars as it could be easily obtained via "client:get_session()".

cartridge/roles/coordinator.lua
```diff
@@ -195,47 +194,53 @@ local function take_control(uri)
         end
 
         if pcall(fiber.status, control_fiber) == 'dead'
```
olegrok, line 196: This condition is always "false". `true/false` ~= 'dead'

cartridge/roles/coordinator.lua
```diff
@@ -195,47 +194,53 @@ local function take_control(uri)
...
-    if vars.connect_fiber == nil then
-        return
-    elseif vars.connect_fiber:status() ~= 'dead' then
-        vars.connect_fiber:cancel()
+    if vars.connect_fiber ~= nil then
+        pcall(fiber.cancel, vars.connect_fiber)
+        vars.connect_fiber = nil
+    end
+
+    if vars.session ~= nil then
```
olegrok, line 235: I suppose that `vars.client:drop_session()` should be enough to drop session. I don't understand such duplication.

cartridge/roles/coordinator.lua
```diff
@@ -247,7 +252,22 @@ local function apply_config(conf, _)
         return true
     end
 
-    if failover_cfg.state_provider ~= 'tarantool' then
+    if failover_cfg.state_provider == 'tarantool' then
+        local params = assert(failover_cfg.tarantool_params)
+
+        if vars.client == nil
```
olegrok, line 258: It would be great if you find a way to generalise such logic. You change it here if ~something~ is changed. I suppose you could organise configuration in a way then `cmpdeeply(new, old)` makes you call client:stop/drop and reconnect then.

cartridge/roles/coordinator.lua
```diff
@@ -298,34 +313,47 @@ local function appoint_leaders(leaders)
...
 
         if server.replicaset_uuid ~= k then
-            return nil, AppointmentError:new('Server "%s" does not belong to replicaset "%s"', v, k)
+            return nil, AppointmentError:new(
+                "Server %q doesn't belong to replicaset %q", v, k
+            )
         end
+    end
+
+    local session = vars.session
```
olegrok, line 335: If you don't want to force reconnect client if session is broken/closed/... add "client.get_session_or_nil" function or smth else with better name.

cartridge/roles/coordinator.lua
```diff
@@ -123,68 +108,76 @@ local function control_loop(conn)
...
             and (decision.immunity < next_moment)
             then
                 next_moment = decision.immunity
             end
         end
 
+        assert(next_moment >= now)
         vars.membership_notification:wait(next_moment - now)
-    until not pcall(fiber.testcancel)
+    until not pcall(fiber.testcancel) or not session:is_locked()
```
olegrok, line 131: Looks strange.

`until not pcall(fiber.testcancel)` is an infinite loop. not session:is_locked() potentially is not infinite. And this mixture looks strange.

Consider a way to move `not session:is_locked()` into loop body and add explicit break.

cartridge/stateboard-client.lua
```diff
@@ -0,0 +1,158 @@
...
+
+    return session.connection:is_connected()
+        and session.lock_acquired
+end
+
+local function is_alive(session)
+    checks('stateboard_session')
+    assert(session.connection ~= nil)
+
+    return session.connection.state == 'initial'
```
olegrok, line 73: image

Let's consider "alive" only as "connection:is_connected"

rosik: Nope. We create session with `wait_connected = false` flag. "Initial" state is appropriate for the session to be counted as alive

olegrok: But there are also "connecting", "auth" states. And there is no difference between "initial" and these states.

cartridge/roles/coordinator.lua
```diff
@@ -123,119 +108,136 @@ local function control_loop(conn)
...
+        then
             break
         end
     until not pcall(fiber.testcancel)
 
-    pcall(conn.close, conn)
-    pcall(fiber.cancel, control_fiber)
-    vars.conn = nil
+    session:drop()
+    pcall(function()
```
olegrok, line 202: pcall(fiber.cancel, control_fiber) looks more appropriate IMO. Feel free to ignore.

PR #731, opened 2020-04-02
Show fullname instead of username in UI auth title

### LINKED ISSUE #726: Add fullname field to graphql authParams
We need to show username (like Tom Chase) instead of username (in cartridge it is email) so I propose to add the fullname field into graphql request. This field we can set to value that the get_user callback returns.

olegrok:
Since "username" is a part of account credentials it's secure
information and should be hidden from public view.
This patch simply changes "username" to "fullname" (if it was
specified) in user's auth params graphql endpoint and auth title
in UI.

Closes #726

- [x] Tests
- [x] Changelog

### DIFF
--- a/cartridge/webui/api-auth.lua
+++ b/cartridge/webui/api-auth.lua
@@ -75,8 +75,21 @@ end
 local function get_auth_params()
     local callbacks = auth.get_callbacks()
     local params = auth.get_params()
+
+    local user, username
+    local session_username = auth.get_session_username()
+    if session_username ~= nil then
+        user = auth.get_user(session_username)
+    end
+
+    if user ~= nil and user.fullname ~= nil then
+        username = user.fullname
+    else
+        username = session_username
+    end
+
     return {
-        username = auth.get_session_username(),
+        username = username,
 
         enabled = params.enabled,
         cookie_max_age = params.cookie_max_age,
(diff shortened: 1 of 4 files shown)


cartridge/webui/api-auth.lua
```diff
@@ -75,8 +75,21 @@ end
...
     local callbacks = auth.get_callbacks()
     local params = auth.get_params()
+
+    local user, username
+    local session_username = auth.get_session_username()
+    if session_username ~= nil then
+        user = auth.get_user(session_username)
+    end
+
+    if user ~= nil and user.fullname ~= nil then
```
mtrempoltsev, line 85: I think we should prefer user.username instead of session_username

olegrok: It was done as it was decided here - https://github.com/tarantool/cartridge/issues/726#issuecomment-607333734

Also seems that "session_username" and "user.username" are identical things, however session_username could be easier extracted from session rather than username from user object.

cartridge/webui/api-auth.lua
```diff
@@ -75,8 +75,21 @@ end
...
+        username = user.fullname
+    else
+        username = session_username
+    end
+
     return {
-        username = auth.get_session_username(),
+        username = username,
```
rosik, line 92:
```suggestion
    local username = auth.get_session_username()
    local user = username and auth.get_user(username)
    return {
        username = user and user.fullname or username,
```

cartridge/webui/api-auth.lua
rosik: It's OK. `user.username == get_session_username()`, `user.fullname == tdg.username`, it's what we discussed yesterday.

PR #762, opened 2020-04-09
Fix rpc_test for macOS

HustonMmmavr:
I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

### DIFF
--- a/test/integration/rpc_test.lua
+++ b/test/integration/rpc_test.lua
@@ -169,11 +169,13 @@ function g.test_routing()
         'myrole', 'void', nil,
         {uri = 'localhost:0'}
     )
-    t.assert_covers({
-        ['"localhost:0": ' .. errno.strerror(errno.ECONNREFUSED)] = true,
-        ['"localhost:0": ' .. errno.strerror(errno.ENETUNREACH)] = true,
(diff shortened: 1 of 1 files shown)


test/integration/rpc_test.lua
```diff
@@ -169,11 +169,13 @@ function g.test_routing()
         'myrole', 'void', nil,
         {uri = 'localhost:0'}
     )
-    t.assert_covers({
-        ['"localhost:0": ' .. errno.strerror(errno.ECONNREFUSED)] = true,
-        ['"localhost:0": ' .. errno.strerror(errno.ENETUNREACH)] = true,
-    }, {[err.err] = true})
-    t.assert_equals(err.class_name, 'NetboxConnectError')
+    t.assert_items_include(
```
olegrok, line 172: I think it should be done in a different way. 

Please don't use nonexistent ports. Simply you could apply this solution - https://github.com/tarantool/cartridge/pull/696/files#diff-3b3a548c9aeec5e1b0b31b02c38cc44aR160 

Or another way it could be two test. One with ":3303" - allowed port and ":0" - nonexistent ports

PR #788, opened 2020-04-16
785 callback config apply

### LINKED ISSUE #24: Implement config modification before validation
In GitLab by @rosik on Nov 27, 2018, 18:31

There should be a way to modify config before validation.

Consider tdg "Upload config" functionality. It generates "ddl" config section and stores it clusterwide.

### LINKED ISSUE #785: Introduce callback on config apply
Currently cartridge.config_patch_clusterwide leads to validate_config and apply_config execution. However, sometimes we have some system sections that could be generated on the base of existing sections. E.g. I have "model" section that generates "DDL" section. "DDL" section is system section in my application and all changes in model section should consider previous state - "DDL" section.

I propose to introduce new callback e.g. `on_config_apply`. Specified function should accept config diff and old config and should return modified config diff that will be applied or error.

Why does it really need. I see only two ways how to "hack" existing restrictions - monkeypatch config upload and create custom config "modifiers" (accepts part of config and calculate diff that I want to apply). Both this approaches interfere an ease of deployment - I can't simply use "ansible-cartridge" that call cartridge.config_patch_clusterwide directly from console.

Let's implement such callback.

```lua
local function on_config_apply(patch, config)
    if config.should_throw_error == true then
        return nil, "Config rejected"
    end
    if patch.user_section ~= nil then
        patch.system_section = calculate(patch.user_section)
    end
    return patch
end

require('cartridge.two-phase').on_patch(my_trigger)

```

HustonMmmavr:
Add callback on config apply

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)
- [x] Documentation

Close #785
Close #24

### DIFF
--- a/cartridge/twophase.lua
+++ b/cartridge/twophase.lua
@@ -34,6 +34,7 @@ yaml.cfg({
 
 vars:new('locks', {})
 vars:new('prepared_config', nil)
+vars:new('on_config_apply', {})
 
 --- Two-phase commit - preparation stage.
 --
@@ -232,6 +233,15 @@ local function _clusterwide(patch)
             end
 
             clusterwide_config_new:set_plaintext(k, yaml.encode(v))
+        end
+    end
+
+    for trigger, _ in pairs(vars.on_config_apply) do
+        local _, err = PatchClusterwideError:pcall(
+            trigger, clusterwide_config_new, clusterwide_config_old
+        )
+        if err ~= nil then
+            return nil, err
         end
     end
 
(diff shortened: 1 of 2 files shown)


cartridge/twophase.lua
```diff
@@ -443,6 +453,44 @@ local function set_schema(schema_yml)
...
+--
+-- If the parameters are `(nil, old_trigger)`, then the old trigger is
+-- deleted.
+--
+-- (**Added** in v2.1.0-1)
+--
+-- @usage
+--    local function update_string_data(conf_new, conf_old)
+--        local old_data = config.get_readonly('data')
+--        conf_new:set_plaintext('data' old_data .. conf_new:get_readonly('data'))
```
olegrok, line 473: I suppose that there is a missing comma in this line

PR #799, opened 2020-04-20
Add console_sock length validation

### LINKED ISSUE #790: Check console_sock length
See https://github.com/tarantool/tarantool/issues/4634

Now there is confused error `Address already in use` because long socket name is cut.

HustonMmmavr:
I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #790

### DIFF
--- a/cartridge.lua
+++ b/cartridge.lua
@@ -585,7 +585,7 @@ local function cfg(opts, box_opts)
         if #unix_port < #opts.console_sock then
             sock:close()
             fio.unlink(unix_port)
-            return nil, CartridgeCfgError:new('Length of console_sock is more than UNIX_PATH_MAX')
+            return nil, CartridgeCfgError:new('Too long console_sock exceeds UNIX_PATH_MAX limit')
         end
     end
 
--- a/cartridge.lua
+++ b/cartridge.lua
@@ -581,7 +581,10 @@ local function cfg(opts, box_opts)
             return nil, err
         end
 
-        if #sock:name().port < #opts.console_sock then
+        local unix_port = sock:name().port
+        if #unix_port < #opts.console_sock then
+            sock:close()
+            fio.unlink(unix_port)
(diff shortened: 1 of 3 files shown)


cartridge.lua
```diff
@@ -576,10 +576,17 @@ local function cfg(opts, box_opts)
...
+        local sock, err = CartridgeCfgError:pcall(console.listen, 'unix/:' .. opts.console_sock)
+        if not sock then
             return nil, err
         end
+
+        local unix_port = sock:name().port
+        if #unix_port < #opts.console_sock then
+            sock:close()
+            fio.unlink(unix_port)
+            return nil, CartridgeCfgError:new('Length of console_sock is more than UNIX_PATH_MAX')
```
rosik, line 588: Too long console_sock exceeds UNIX_PATH_MAX limit

test/unit/cfg_errors_test.lua
```diff
@@ -234,6 +234,19 @@ g.test_auth_backend = function()
...
+    g.mock_membership()
+    check_error('Length of console_sock is more than UNIX_PATH_MAX',
+        cartridge.cfg, {
+            workdir = '/tmp',
+            advertise_uri = 'unused:0',
+            http_enabled = false,
+            roles = {},
+            console_sock = string.format('%s/%s.sock', g.tempdir, ('a'):rep(108))
+        }
+    )
```
rosik, line 247: check that unlink works

dokshina:
What about notify socket?
https://github.com/tarantool/cartridge/blob/master/cartridge.lua#L585

rosik:
> What about notify socket?
> https://github.com/tarantool/cartridge/blob/master/cartridge.lua#L585

1. It's not created by cartridge, and it's not his concern.
1. It doesn't make sense in modern tarantool.
1. User can always pass non-existent path even if it's short enough. It's not a fatal error.

So I think it's not necessary to do anything similar about `notify_socket`.

PR #808, opened 2020-04-22
Fix leaking sockets in remote control

rosik:
Consider the case when cartridge instance recovers from snapshot.
It calls `remote_control.bind()` and accepts new connections, but doesn't
reply anything until recovery finishes. It may take some time.

If a client connection times out, the socket remains in CLOSE_WAIT state.
It's not GC'ed until the handler returns and socket is closed from
the server side explicitly.

If a client tries to reconnect afterwards, it results in fd leak, and
after some time the instance will likely get "too many open files" error.
We've met similar problem [here](https://github.com/tarantool/tarantool/issues/4910).

After this patch tcp handler doesn't retain closed sockets forever.

I didn't forget about

- [x] Tests
- [x] Changelog

### DIFF
--- a/cartridge/remote-control.lua
+++ b/cartridge/remote-control.lua
@@ -308,9 +308,29 @@ local function rc_handle(s)
     vars.handlers[s] = fiber.self()
     s._client_user = 'guest'
     s._client_salt = salt
-    if not vars.accept then
-        vars.accept_cond:wait()
+
+    -- When a client reconnects before the server started accepting
+    -- requests it's reasonable to avoid waiting for one-second timeout.
+    -- We wake up all the handlers so they could check if their
+    -- sockets were closed.
+    vars.accept_cond:broadcast()
+
+    while not vars.accept do
+        vars.accept_cond:wait(1)
+        fiber.testcancel()
+
+        -- Socket can become readable in two cases:
+        -- 1. It was closed by the client. On the server side it stays
+        --    in CLOSE_WAIT state. The server should close it and
+        --    release asap.
+        -- 2. Server received data even before it has sent the greeting.
+        --    This case doesn't conform to Tarantool iproto protocol.
+        if s:readable(0) then
+            vars.handlers[s] = nil
+            return
+        end
     end
+
     s:write(greeting)
 
     while vars.handlers[s] do
@@ -398,8 +418,7 @@ local function stop()
 
     vars.server:close()
     vars.server = nil
-    vars.accept = nil
-    vars.accept_cond = nil
+    vars.accept = false
 end
 
 --- Explicitly drop all established connections.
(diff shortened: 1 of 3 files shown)


cartridge/remote-control.lua
```diff
@@ -308,9 +308,28 @@ local function rc_handle(s)
...
+        vars.accept_cond:wait(1)
+
+        -- Socket can become readable in two cases:
+        -- 1. It was closed by the client. The socket remains in
+        --    CLOSE_WAIT state and the server should close it and
+        --    release asap.
+        -- 2. Server received data even before it has sent the greeting.
+        --    This case doesn't conform to Tarantool iproto protocol.
+        if s:readable(0) then
+            vars.handlers[s] = nil
```
filonenko-mikhail, line 328: Dont you want to close client socket here?

filonenko-mikhail: socket.tcp_server makes it

test/unit/remote_control_test.lua
```diff
@@ -813,3 +813,24 @@ function g.test_reconnect()
...
     end)
 end
+
+function g.test_gc()
+    local ok, err = remote_control.bind('127.0.0.1', 13301)
+    t.assert_equals(err, nil)
+    t.assert_equals(ok, true)
+
+    local RLIMIT_NOFILE = tonumber(io.popen('ulimit -n'):read('*all'))
+    require('log').info('RLIMIT_NOFILE: %d', RLIMIT_NOFILE)
```
olegrok, line 823: Seems this is a debug line

test/unit/remote_control_test.lua
```diff
@@ -813,3 +813,24 @@ function g.test_reconnect()
         })
     end)
 end
+
+function g.test_gc()
```
olegrok, line 817: test_<socket/fd>_gc?

test/unit/remote_control_test.lua
rosik: Well yes, but actually no. (But yes).
Anyway, it's helpful in CI.

rosik: Yet ulimit didn't work in CI, so I refactored the test

test/unit/remote_control_test.lua
```diff
@@ -813,3 +814,34 @@ function g.test_reconnect()
...
+        return #fio.listdir(procfd) - 1
+    end
+
+    local initial_fdcnt = fdcnt()
+
+    for i = 1, 16 do
+        local s = socket.tcp_connect('127.0.0.1', 13301)
+        t.assert(s, string.format('socket #%s: %s', i, errno.strerror()))
+        s:close()
+        require('fiber').sleep(0)
```
olegrok, line 836: Why can't you require fiber globally to the whole file?

PR #848, opened 2020-05-18
Enable stateboard in development

rosik:
These changes are necessary to easily configure stateful failover in development environment

I didn't forget about

- [x] Tests (unnecessary)
- [x] Changelog (unnecessary)
- [x] Documentation

### DIFF
--- a/instances.yml
+++ b/instances.yml
@@ -25,3 +25,8 @@ cartridge.srv-5:
     workdir: ./dev/3305
     advertise_uri: localhost:3305
     http_port: 8085
+
+cartridge-stateboard:
+    workdir: ./dev/stateboard
+    password: qwerty
+    listen: 4401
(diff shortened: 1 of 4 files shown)


instances.yml
```diff
@@ -25,3 +25,8 @@ cartridge.srv-5:
     workdir: ./dev/3305
     advertise_uri: localhost:3305
     http_port: 8085
+
+cartridge-stateboard:
+    workdir: ./dev/stateboard
+    password: qwerty
+    listen: 4401
```
dokshina, line 32: Why not `localhost:4401`?

stateboard.init.lua
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env tarantool
+
+require('cartridge.stateboard').cfg()
```
dokshina, line 3: What about `setsearchroot` for Tarantool 1.10?

rosik: It's only necessary for cartridge-based projects packaging, which is never used for bare cartridge. In `.rock/bin` it's already prefixed appropriately.

PR #873, opened 2020-06-02
WIP: roles: introduce metrics role

### LINKED ISSUE #92: Expose default metrics from admin httpd
In GitLab by @sharonovd on Apr 29, 2019, 11:23

Utilize default metrics from `metrics` for the purpose.

That would make operations life easier. The only trick would be to leave them extenadble.

vasiliy-t:
WIP: lets agree on configuration format and I'll complete this pr with tests, expose metrics as a service in service_registry 

proposed configuration format:

```yaml
metrics:
      export:
        - path: "/metrics/json"
          format: "json"
        - path: "/metrics/prom"
          format: "prometheus"
      collect:
        default:
```

where 

- ```metrics``` is a top level section name
- ```export``` is exporter configuration, e.g. [1] is a way to enable json metrics via http endpoint /metrics/json
- ```collect``` is a collectors configuration, e.g. ```default``` means that enable_json_metrics will be called, required for further extensions, for example upcoming metrics ```psutils``` 

What has been done? New builtin role to expose metrics.

Why? Lack of standard and simple way to expose metrics

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Close #92

### DIFF
--- a/cartridge-scm-1.rockspec
+++ b/cartridge-scm-1.rockspec
@@ -14,6 +14,7 @@ dependencies = {
     'vshard == 0.1.15-1',
     'membership == 2.2.0-1',
     'frontend-core == 6.5.1-1',
+    'metrics == 0.2.0'
 }
 
 external_dependencies = {
--- a/cartridge/roles/instrumentation/metrics.lua
+++ b/cartridge/roles/instrumentation/metrics.lua
@@ -1,0 +1,66 @@
+local cartridge = require("cartridge")
+local argparse = require("cartridge.argparse")
+local metrics = require("metrics")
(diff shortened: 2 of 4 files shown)


printercu:
If you decide to proceed with using clusterwide config for configuring this role, I suggest to provide api to set default config from the source code. This makes it possible to use this role in applications that don't require any extra configuration to work.

Also maybe it's worth to make this role hidden or even permanent by default.

vasiliy-t:
closing this in favour of https://github.com/tarantool/metrics/pull/74

PR #887, opened 2020-06-08
feedback: include rocks versions from manifest into report

RunsFor:
Before only cartridge version was detected and reported by feedback daemon.

This patch reads manifest file, parses it and include rocks versions in it into feedback report

Example:

    {
      "cartridge": "scm-1",
      "lulpeg": "0.1.2-1",
      "cartridge-cli": "scm-1",
      "membership": "2.2.0-1",
      "frontend-core": "6.5.1-1",
      "errors": "2.1.3-1",
      "vshard": "0.1.15-1",
      "http": "1.1.0-1",
      "ddl": "1.1.0-1",
      "checks": "3.0.1-1"
    }

- [x] Tests
- [x] Changelog
- [ ] Documentation

Relates to https://github.com/tarantool/tarantool/issues/4391

### DIFF
--- a/cartridge/feedback.lua
+++ b/cartridge/feedback.lua
@@ -1,3 +1,4 @@
+local fio = require('fio')
 local log   = require('log')
 local json  = require('json')
 local fiber = require('fiber')
@@ -12,7 +13,7 @@ end
 
 local _daemon_mt = getmetatable(daemon)
 
--- copy-pased from tarantool-1.10.3-124-g3b1e75ece
+-- copy-pasted from tarantool-1.10.3-124-g3b1e75ece
 local function get_fiber_id(f)
     local fid = 0
     if f ~= nil and f:status() ~= "dead" then
@@ -21,7 +22,7 @@ local function get_fiber_id(f)
     return fid
 end
 
--- copy-pased from tarantool-1.10.3-124-g3b1e75ece
+-- copy-pasted from tarantool-1.10.3-124-g3b1e75ece
 local function feedback_loop(self)
     fiber.name(PREFIX, { truncate = true })
 
@@ -38,7 +39,7 @@ local function feedback_loop(self)
     self.shutdown:put("stopped")
 end
 
--- copy-pased from tarantool-1.10.3-124-g3b1e75ece
+-- copy-pasted from tarantool-1.10.3-124-g3b1e75ece
 local function guard_loop(self)
     fiber.name(string.format("guard of %s", PREFIX), {truncate=true})
 
@@ -57,7 +58,7 @@ local function guard_loop(self)
     self.shutdown:put("stopped")
 end
 
--- copy-pased from tarantool-1.10.3-124-g3b1e75ece
+-- copy-pasted from tarantool-1.10.3-124-g3b1e75ece
 local function start(self)
     self:stop()
     if self.enabled then
@@ -68,7 +69,7 @@ local function start(self)
     log.verbose("%s started", PREFIX)
 end
 
--- copy-pased from tarantool-1.10.3-124-g3b1e75ece
+-- copy-pasted from tarantool-1.10.3-124-g3b1e75ece
 local function stop(self)
     if (get_fiber_id(self.guard) ~= 0) then
         self.guard:cancel()
@@ -111,6 +112,31 @@ function _daemon_mt.__index.generate_feedback()
     end
 
     feedback.rocks['cartridge'] = require('cartridge').VERSION
+
+    local searchroot
+    if package.searchroot ~= nil then
+        searchroot = package.searchroot()
+    else
+        searchroot = fio.abspath(fio.dirname(arg[0]))
+    end
+
+    local manifest = {}
+    local manifest_path = fio.pathjoin(searchroot,
+        '.rocks/share/tarantool/rocks/manifest'
+    )
+
+    if not pcall(loadfile(manifest_path, 't', manifest))
+    or type(manifest.dependencies) ~= 'table'
+    then
+        return feedback
+    end
+
+    for rock_name, versions in pairs(manifest.dependencies) do
+        if feedback.rocks[rock_name] == nil then
+            feedback.rocks[rock_name] = next(versions)
+        end
+    end
+
     return feedback
 end
 
--- a/test/integration/feedback_test.lua
+++ b/test/integration/feedback_test.lua
@@ -51,6 +51,7 @@ g.before_all = function()
     })
     g.cluster:start()
 end
+
 g.after_all = function()
     g.httpd:stop()
     g.cluster:stop()
@@ -64,4 +65,5 @@ function g.test_feedback()
     t.assert_equals(msg.server_id, helpers.uuid('a', 'a', 1))
     t.assert_equals(msg.cluster_id, helpers.uuid('a'))
     t.assert_type(msg.rocks.cartridge, 'string')
+    t.assert_type(msg.rocks.http, 'string')
 end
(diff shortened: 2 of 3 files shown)


artur-barsegyan:
As I remember, a bundled application has a special manifest file with a list of installed versions of rocks. `cartridge-cli` can read it. 

**In general:** now, we don't have too many live Cartridge instances for representative analysis. It's nearly 3,5 % of total live Tarantool instances at all. What the main problem to track modules from core?

cartridge/feedback.lua
```diff
@@ -1,18 +1,68 @@
...
+end
+
+local function parse_manifest(filename)
+    local manifest = parse_lua_file(filename)
+    local deps = {}
+
+    for module, versions in pairs(manifest.dependencies or {}) do
+        for version, _ in pairs(versions) do
+            if deps[module] ~= nil then
+                break -- Decide how to deal with duplicates
```
rosik, line 24: Do you have any ideas?

cartridge/feedback.lua
```diff
@@ -1,18 +1,68 @@
...
+local function fill_in_rocks(feedback)
+    feedback.rocks = feedback.rocks or {}
+
+    -- tarantool 1.10 doesn't support package.searchroot
+    -- so we are basically guessing by picking directory
+    -- of the init script. It is a default scenario
+    local app_dir = package.searchroot and package.searchroot() or fio.abspath(fio.dirname(arg[0]))
+    local manifest_path = fio.pathjoin(app_dir, '.rocks/share/tarantool/rocks/manifest')
+
+    if not parsed_manifest then
```
rosik, line 44: This code isn't friendly to hot-reload. According to recent discussions in #842  I suggest to prepare for it.

cartridge/feedback.lua
```diff
@@ -1,18 +1,68 @@
...
+        if fio.path.is_file(manifest_path) then
+            parsed_manifest = parse_manifest(manifest_path)
```
rosik, line 46: The function is used only once. What do you think about inlining it?

cartridge/feedback.lua
```diff
@@ -1,18 +1,68 @@
...
+        end
+    end
+
+    local rocks = parsed_manifest or {}
```
rosik, line 50: Variables and functions naming should be revised all across the PR. How could you assign `parsed_manifest` to `rocks`? Does it hold rocks of manifest?

rosik: `or {}` seems to be unnecessary. `parsed_manifest` can't be nil afaics.

test/integration/feedback_test.lua
```diff
@@ -64,4 +65,5 @@ function g.test_feedback()
     t.assert_equals(msg.server_id, helpers.uuid('a', 'a', 1))
     t.assert_equals(msg.cluster_id, helpers.uuid('a'))
     t.assert_type(msg.rocks.cartridge, 'string')
+    t.assert_type(msg.rocks.http, 'string')
```
rosik, line 68: This test isn't correct. Cartridge dev layout doesn't conform to layout of regular project, but it still passes.

cartridge/feedback.lua
```diff
@@ -106,11 +156,8 @@ function _daemon_mt.__index.generate_feedback()
         return nil, err
     end
 
-    if feedback.rocks == nil then
-        feedback.rocks = {}
-    end
+    fill_in_rocks(feedback)
```
rosik, line 159: What's the purpose of splitting this logic to a separate function?

cartridge/feedback.lua
```diff
@@ -1,18 +1,68 @@
...
+    local state = {}
+    local eval = loadfile(name, 't', state)
+    pcall(eval)
+
+    return state
+end
+
+local function parse_manifest(filename)
+    local manifest = parse_lua_file(filename)
+    local deps = {}
```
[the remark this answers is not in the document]
RunsFor: For me first version is more readable, but I won't insist.

cartridge/feedback.lua
RunsFor: I'm not sure we can anything with it by just parsing `manifest` file. I think, we may just assume, there is consistent `.rocks/` directory and just take the first version.

More reliable approach needs to examine `.rocks/` structure.

cartridge/feedback.lua
RunsFor: I'm more on the "split for readability" side here as well as with `parse_lua_file` function. If `cartridge`'s code style prefer inlining, I'm okay with that. Do you want me to inline it?

cartridge/feedback.lua
RunsFor: See [comment](https://github.com/tarantool/cartridge/pull/887#discussion_r441490797) above. Let's decide, which is more consistent for `cartridge` project.

test/integration/feedback_test.lua
RunsFor: What do you mean?

I needed to assure, that rocks from the manifest will appear in the feedback report. I can't check exact versions and underlaying manifest to a random location would be tedious. So I decided to check for the present of `http` module in report, since it is one of the most stable dependencies for cartridge and test are supposed to run under cartridge environment.

cartridge/feedback.lua
RunsFor: Agree. Thanks for pointing it out.

I've removed confusing `parsed_manifest` variable, and put everything under cached branch `if not rocks then ... end`

RunsFor: > or {} seems to be unnecessary. parsed_manifest can't be nil afaics.

If manifest happen to be not a file, rocks may still be nil

cartridge/feedback.lua
```diff
@@ -1,18 +1,68 @@
...
+
+local parsed_manifest = false
+
+local function fill_in_rocks(feedback)
+    feedback.rocks = feedback.rocks or {}
+
+    -- tarantool 1.10 doesn't support package.searchroot
+    -- so we are basically guessing by picking directory
+    -- of the init script. It is a default scenario
+    local app_dir = package.searchroot and package.searchroot() or fio.abspath(fio.dirname(arg[0]))
```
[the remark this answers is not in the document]
RunsFor: Why so?

cartridge/feedback.lua
```diff
@@ -1,18 +1,68 @@
...
+                break -- Decide how to deal with duplicates
+            end
+            deps[module] = version
+        end
+    end
+
+    return deps
+end
+
+local parsed_manifest = false
```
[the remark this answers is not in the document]
RunsFor: Seems reasonable. fixed

cartridge/feedback.lua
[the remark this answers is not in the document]
RunsFor: Or do you want to simple remove calling `package.searchroot()` here and leave `arg[0]`? So you are saying that `.rocks` directory will more like to be in `arg[0]` then in `package.searchroot()`?

cartridge/feedback.lua
RunsFor: Add file modification time based caching for manifest. Also it will be invalidated if manifest when removed

cartridge/feedback.lua
RunsFor: Decided to take only first version, since it is difficult to detect which version is actually used and requires parsing a file tree

cartridge/feedback.lua
RunsFor: Decided not to deal with caching since feedback daemon runs only once in hour.

cartridge/feedback.lua
RunsFor: Decided to inline it

cartridge/feedback.lua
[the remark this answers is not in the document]
RunsFor: Decided to fallback to arg[0] only when `package.searchroot` is not available

PR #919, opened 2020-06-29
graphql: fix error messages for output types mismatch

### LINKED ISSUE #891: Non-informative error in case when boolean returned instead of array
https://github.com/tarantool/cartridge/blob/502cf09992dab4365349ecbc07cf8ab9a8094b64/cartridge/graphql/execute.lua#L200

```log
WARNING  localhost:11001:server.py:40 2020-06-09 11:49:30.550 [10501] main/182/http/127.0.0.1:63423 init.lua:114 E> [a1b04dd3-32b0-4325-98f8-e32255fa7712] Graphql execution failed: graphql internal error: ...tdg/.rocks/share/tarantool/cartridge/graphql/execute.lua:201: attempt to concatenate field 'name' (a nil value)
WARNING  localhost:11001:server.py:40 stack traceback:
WARNING  localhost:11001:server.py:40 	...tdg/.rocks/share/tarantool/cartridge/graphql/execute.lua:307: in function 'evaluateSelections'
WARNING  localhost:11001:server.py:40 	...tdg/.rocks/share/tarantool/cartridge/graphql/execute.lua:217: in function 'completeValue'
WARNING  localhost:11001:server.py:40 	...tdg/.rocks/share/tarantool/cartridge/graphql/execute.lua:178: in function 'getFieldEntry'
WARNING  localhost:11001:server.py:40 	...tdg/.rocks/share/tarantool/cartridge/graphql/execute.lua:307: in function 'evaluateSelections'
WARNING  localhost:11001:server.py:40 	...tdg/.rocks/share/tarantool/cartridge/graphql/execute.lua:217: in function 'getFieldEntry'
WARNING  localhost:11001:server.py:40 	...tdg/.rocks/share/tarantool/cartridge/graphql/execute.lua:307: in function 'evaluateSelections'
WARNING  localhost:11001:server.py:40 	...tdg/.rocks/share/tarantool/cartridge/graphql/execute.lua:217: in function 'completeValue'
WARNING  localhost:11001:server.py:40 	...tdg/.rocks/share/tarantool/cartridge/graphql/execute.lua:206: in function 'getFieldEntry'
WARNING  localhost:11001:server.py:40 	...tdg/.rocks/share/tarantool/cartridge/graphql/execute.lua:307: in function <...tdg/.rocks/share/tarantool/cartridge/graphql/execute.lua:301>
WARNING  localhost:11001:server.py:40 	...
```

olegrok:
This patch fixes two problems.

1. When we return wrong value for List of nonNull type. nil dereference happened because error handler didn't consider nonNullable wrapper. `innerType.name` for `Type` is `"Type"` and `nil` for `Type!` (expected "Type" or "NonNull(Type)"). (https://github.com/tarantool/cartridge/pull/919/files#diff-926c2702f04ab81baa008eec661b481eR200)

1. Scalar could be passed instead of Object. This patch adds a check that Object is "table". (https://github.com/tarantool/cartridge/pull/919/files#diff-926c2702f04ab81baa008eec661b481eR218)

Closes #891

### DIFF
--- a/cartridge/graphql/execute.lua
+++ b/cartridge/graphql/execute.lua
@@ -194,13 +194,16 @@ local function completeValue(fieldType, result, subSelections, context, opts)
   end
 
   if fieldTypeName == 'List' then
-    local innerType = fieldType.ofType
-
-    if type(result) ~= 'table' then
-      local message = ('Expected a table for %q list, got %q'):format(util.getTypeName(innerType), type(result))
+    if not util.is_array(result) then
+      local resultType = type(result)
+      if resultType == 'table' then
+        resultType = 'map'
+      end
+      local message = ('Expected %q to be an "array", got %q'):format(fieldName, resultType)
       error(message)
     end
 
+    local innerType = fieldType.ofType
     local values = {}
     for i, value in ipairs(result) do
       values[i] = completeValue(innerType, value, subSelections, context)
@@ -215,7 +218,7 @@ local function completeValue(fieldType, result, subSelections, context, opts)
 
   if fieldTypeName == 'Object' then
     if type(result) ~= 'table' then
-      local message = ('Expected a table for %q object, got %q'):format(util.getTypeName(fieldType), type(result))
+      local message = ('Expected %q to be a "map", got %q'):format(fieldName, type(result))
       error(message)
     end
     local completed = evaluateSelections(fieldType, result, subSelections, context)
--- a/cartridge/graphql/execute.lua
+++ b/cartridge/graphql/execute.lua
@@ -72,11 +72,6 @@ local function mergeSelectionSets(fields)
 end
 
 local function defaultResolver(object, _, info)
-  if type(object) ~= 'table' then
-    local objectType = info.fieldASTs[1].kind
-    local name = info.fieldASTs[1].name.value
-    error(('Expected a table for %s %q'):format(objectType, name))
-  end
   return object[info.fieldASTs[1].name.value]
 end
 
@@ -202,10 +197,8 @@ local function completeValue(fieldType, result, subSelections, context, opts)
     local innerType = fieldType.ofType
 
     if type(result) ~= 'table' then
-      if innerType.__type == 'NonNull' then
-        innerType = innerType.ofType
-      end
-      error('Expected a table for ' .. innerType.name .. ' list')
+      local message = ('Expected a table for %q list, got %q'):format(util.getTypeName(innerType), type(result))
+      error(message)
     end
 
     local values = {}
@@ -221,6 +214,10 @@ local function completeValue(fieldType, result, subSelections, context, opts)
   end
 
   if fieldTypeName == 'Object' then
+    if type(result) ~= 'table' then
+      local message = ('Expected a table for %q object, got %q'):format(util.getTypeName(fieldType), type(result))
+      error(message)
+    end
     local completed = evaluateSelections(fieldType, result, subSelections, context)
     setmetatable(completed, serializemap)
     return completed
--- a/cartridge/graphql/execute.lua
+++ b/cartridge/graphql/execute.lua
@@ -72,6 +72,11 @@ local function mergeSelectionSets(fields)
 end
 
 local function defaultResolver(object, _, info)
+  if type(object) ~= 'table' then
+    local objectType = info.fieldASTs[1].kind
+    local name = info.fieldASTs[1].name.value
+    error(('Expected a table for %s %q'):format(objectType, name))
+  end
   return object[info.fieldASTs[1].name.value]
 end
 
@@ -197,6 +202,9 @@ local function completeValue(fieldType, result, subSelections, context, opts)
     local innerType = fieldType.ofType
 
     if type(result) ~= 'table' then
+      if innerType.__type == 'NonNull' then
+        innerType = innerType.ofType
+      end
       error('Expected a table for ' .. innerType.name .. ' list')
     end
 
--- a/test/integration/graphql_test.lua
+++ b/test/integration/graphql_test.lua
@@ -740,68 +740,53 @@ g.test_output_type_mismatch_error = function()
     server.net_box:eval([[
         local fiber = require('fiber')
         package.loaded['test'] = {}
-        package.loaded['test']['expected_list'] = function(_, args)
-            if args.type == 'boolean' then
-                return {{value = {true}}}
-            elseif args.type == 'boolean_list' then
-                return {{value = {{value = true}}}}
-            elseif args.type == 'string' then
-                return {{value = {'Hello'}}}
-            elseif args.type == 'number' then
-                return 123
-            elseif args.type == 'cdata' then
-                return 123LL
-            elseif args.type == 'userdata' then
-                return fiber.create(function() end)
-            elseif args.type == 'object' then
-                return {a = 'b'}
-            end
-            return {{value = {{value = {1}}}}, {value = {{value = {2}}}}}
+        package.loaded['test']['callback'] = function(_, args)
+            return true
         end
 
         local graphql = require('cartridge.graphql')
         local types = require('cartridge.graphql.types')
 
-        local nested_type = types.object({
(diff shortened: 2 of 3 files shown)


test/integration/graphql_test.lua
```diff
@@ -687,6 +736,174 @@ g.test_custom_type_scalar_variables = function()
...
+    )
+
+    t.assert_error_msg_equals('Expected a table for field "value"', function()
+        return server:graphql({
+            query = [[
+                query($type: String) {
+                    expected_list(type: $type) { value { value } }
+                }
+            ]],
+        variables = {type = 'boolean'}}
```
RunsFor, line 809: You are testing error messages here. Declaring variables looks unnecessary too. Same for the following tests

test/integration/graphql_test.lua
```diff
@@ -343,6 +343,55 @@ g.test_enum = function()
...
+            name = 'simple_object',
+            fields = {
+                value = simple_enum,
+            }
+        })
+
+        graphql.add_callback({
+            name = 'simple_enum_output',
+            args = {
+                error = types.string,
```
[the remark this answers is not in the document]
olegrok: Yes, thanks. Removed.

test/integration/graphql_test.lua
olegrok: Variable value changes return value. And I see no bad to use graphql variables in such case.

RunsFor: It's not bad, it's just not necessary. You can put this value inside `type` argument without using variables.

RunsFor:
Could you please explain why error messages are so different?

For `string` argument it is:

```
'Expected a table for field "value"'
```

For `cdata`:

```
'Expected a table for ObjectWithValue list'
```

I would rather expect a following format of the error message:

```
Expected <expected type> for field <field name>, got <value> of type <value type>
```

olegrok:
> Could you please explain why error messages are so different?

It's just different validation branches. But I agree, I'll try to unify error message.

RunsFor:
Sorry, I didn't fully understand your changes.

In GraphQL spec there are so called **result coercion** rules. Basically you execute a resolver, then apply one of result coercion rules. If coercion not possible, you raise an error, otherwise return coerced value.

I don't see any result coercion in the code.

olegrok:
> Sorry, I didn't fully understand your changes.

Well, this patch doesn't have a target to be more compatible with GraphQL spec.
There are some places (two of them I want to fix in this patch) where type mismatch leads to errors "attempt to index nil value". Such things happen when you try to return some scalar value instead of list (it could not be coerced anyway) and another case when error happen but you use nonNullable type and just forget to retrieve original type - nonNull doesn't have "name" field.

In short this patch tries to eliminate *Lua* errors (attempt to index a nil value) but not GraphQL.

RunsFor:
> Such things happen when you try to return some scalar value instead of list (it could not be coerced anyway) and another case when error happen but you use nonNullable type and just forget to retrieve original type - nonNull doesn't have "name" field.

Ok, then. I suppose you may extract this validating logic from resolvers and check constraints after resolving and before completing value.

I believe it is possible to return scalar types from custom resolvers.

test/integration/graphql_test.lua
olegrok: Removed

olegrok:
> > Such things happen when you try to return some scalar value instead of list (it could not be coerced anyway) and another case when error happen but you use nonNullable type and just forget to retrieve original type - nonNull doesn't have "name" field.
> 
> Ok, then. I suppose you may extract this validating logic from resolvers and check constraints after resolving and before completing value.
> 
> I believe it is possible to return scalar types from custom resolvers.

@RunsFor

I've reworked my patch:
- Improve error message verbosity - `Expected a table for "Type" list/object, got "lua-type"`
- I'm not sure that we should extract something. Validation logic is strictly tied with completion logic. Scalar values were prohibited to be returned instead of lists - I've saved such behaviour, just add nullability consideration. And I've made the same for "Object" type. Now I suppose it looks more appropriate.

cartridge/graphql/execute.lua
```diff
@@ -194,12 +194,12 @@ local function completeValue(fieldType, result, subSelections, context, opts)
   end
 
   if fieldTypeName == 'List' then
-    local innerType = fieldType.ofType
-
-    if type(result) ~= 'table' then
-      error('Expected a table for ' .. innerType.name .. ' list')
+    if type(result) ~= 'table' or not util.is_array(result) then
```
RunsFor, line 197: maybe simplify to `if not util.is_array(result) then`?

cartridge/graphql/execute.lua
```diff
@@ -213,6 +213,10 @@ local function completeValue(fieldType, result, subSelections, context, opts)
   end
 
   if fieldTypeName == 'Object' then
+    if type(result) ~= 'table' then
```
RunsFor, line 216: And `if not util.is_map(result)` here?

cartridge/graphql/execute.lua
```diff
@@ -213,6 +213,10 @@ local function completeValue(fieldType, result, subSelections, context, opts)
...
+      local message = ('Expected %q to be a "table", got %q'):format(fieldName, type(result))
```
RunsFor, line 217: Also, to be consistent, I would change to pattern `'Expected %q to be a "map", got %q'`

cartridge/graphql/execute.lua
olegrok: It makes us to patch is_array function a bit. But it seems to be ok.

cartridge/graphql/execute.lua
olegrok: It's redundant I think. Because you should declare nested fields for map. And array won't have them. In contrast map was interpreted as an empty array thus is_array check makes sense.

test/integration/graphql_test.lua
```diff
@@ -687,6 +734,105 @@ g.test_custom_type_scalar_variables = function()
...
+    t.assert_error_msg_equals('Expected "expected_list" to be an "array", got "boolean"', function()
+        return server:graphql({
+            query = [[
+                query {
+                    expected_list
+                }
+            ]]})
+    end)
+
+    t.assert_error_msg_equals('Expected "expected_list_with_nested" to be an "array", got "table"', function()
```
RunsFor, line 826: When using `array`/`map` notation, "table" become confusing.

cartridge/graphql/execute.lua
olegrok: It's redundant I think. Because you should declare nested fields for map. And array won't have them. In contrast map was interpreted as an empty array thus is_array check makes sense.

cartridge/graphql/execute.lua
olegrok: Agree, fixed

test/integration/graphql_test.lua
olegrok: Do you have any alternatives?

cartridge/graphql/execute.lua
RunsFor: I don't get it. Object queries with no selections are not allowed in graphql. So this check may be done. If at least one field is not nullable, then resolver has to return an object with at least one string key.

Anyway, on the other hand there is a particular case, when an object has all fields nullable. There will be empty map indeed, which is indistinguishable with an empty array in lua. So lets leave this peculiarity as is for now.

test/integration/graphql_test.lua
RunsFor: What about adding additional check - when `result` is a table, check that it has no string keys via `type(next(result))`. If so, print "map" in the error message. If not, just print result's type.

olegrok: Seems could be simply done with:
```lua
local resultType = type(result)
if resultType == 'table' then
    resultType = 'map'
end
```

olegrok: Finally, `Expected "expected_list_with_nested" to be an "array", got "map"`

PR #920, opened 2020-06-29
graphql: consider default values for graphql execution

### LINKED ISSUE #866: graphql: default variables are not working
Two screenshot below should be self-explanatory. First query without default variable:

Screenshot 2020-06-02 at 11 49 31

Second with uuid default variable:

Screenshot 2020-06-02 at 11 50 17

One server record were expected to be in response

olegrok:
Closes #866

### DIFF
--- a/test/integration/graphql_test.lua
+++ b/test/integration/graphql_test.lua
@@ -779,3 +779,51 @@ function g.test_middleware()
         }
     )
 end
+
+g.test_default_values = function()
+    local server = cluster.main_server
+
+    server.net_box:eval([[
+        package.loaded['test'] = package.loaded['test'] or {}
+        package.loaded['test']['test_default_value'] = function(_, args)
+            if args.arg == nil then
+                return 'nil'
+            end
+            return args.arg
+        end
+
+        local graphql = require('cartridge.graphql')
+        local types = require('cartridge.graphql.types')
+
+        graphql.add_callback({
+            name = 'test_default_value',
+            args = {
+                arg = types.string,
+            },
+            kind = types.string,
+            callback = 'test.test_default_value',
+        })
+    ]])
+
+    t.assert_equals(
+        server:graphql({
+            query = [[
+                query($arg: String = "default_value") {
+                    test_default_value(arg: $arg)
+                }
+            ]],
+        variables = {}}
+        ).data.test_default_value, 'default_value'
+    )
+
+    t.assert_equals(
+        server:graphql({
+            query = [[
+                query($arg: String = "default_value") {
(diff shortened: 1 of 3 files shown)


test/integration/graphql_test.lua
```diff
@@ -779,3 +779,37 @@ function g.test_middleware()
...
+    ]])
+
+    t.assert_equals(
+        server:graphql({
+            query = [[
+                query($arg: String = "default_value") {
+                    test_default_value(arg: $arg)
+                }
+            ]],
+        variables = {arg = box.NULL}}
```
RunsFor, line 812: I believe, if there is an explicit `null` value, it should override default value. I suggest split this into 2 tests:

1. `variables = {}` - returns `default value`
2. `variables = {arg = box.NULL}` returns `nil`

RunsFor: Also, consider a case, when argument is nonNull. Query should return an error if default value is provided, but variable has been set to `box.NULL`

olegrok: I agree with you - https://github.com/graphql/graphql-spec/pull/418 (seems it's covered in spec).

So, I'll try to fix it.

olegrok: @RunsFor 

I've updated my MR. Since that default arrays and objects are supported. And there is a difference between explicitly passed "null" and just omitted argument. However, I've noticed that nonNull arguments can't have default values -  https://github.com/tarantool/cartridge/blob/master/cartridge/graphql/rules.lua#L411

test/integration/graphql_test.lua
```diff
@@ -779,3 +779,134 @@ function g.test_middleware()
...
+
+    t.assert_equals(
+        server:graphql({
+            query = [[
+                query($arg: default_input_object = {field: "default_value"}) {
+                    test_default_object(arg: $arg)
+                }
+            ]],
+        variables = {arg = box.NULL}}
+        ).data.test_default_object, 'nil'
```
RunsFor, line 910: Are you sure? I suppose it should return nothing (`nil`), not string `"nil"`

olegrok: Cartridge doesn't allow to return "nil" as a value. I return "nil" as string in my functions in such case - https://github.com/tarantool/cartridge/pull/920/files#diff-cf7848a656b50a8b4ce2cc926bef5205R790

olegrok: There is a root of such workaroundhttps://github.com/tarantool/cartridge/blob/1f4a8e026fefa2d8434bca4a2a2cde29750d59a9/cartridge/graphql.lua#L304

PR #921, opened 2020-06-29
Implement wait_lsn

### LINKED ISSUE #904: [4pt] Implement wait_lsn function
Prerequisite for #749

Steap2448:
Wait for `box.info.vclock[id]` to reach desired LSN

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary, private API)
- [x] Documentation

Close #904

### DIFF
--- a/cartridge/utils.lua
+++ b/cartridge/utils.lua
@@ -1,6 +1,7 @@
 local fio = require('fio')
 local ffi = require('ffi')
 local errno = require('errno')
+local fiber = require('fiber')
 local checks = require('checks')
 local errors = require('errors')
 
@@ -205,7 +206,7 @@ local function set_readonly(tbl, ro)
     return tbl
 end
 
---- Return true if we run under systemd
+--- Return true if we run under systemd.
 -- systemd detection based on http://unix.stackexchange.com/a/164092
 local function under_systemd()
     local rv = os.execute("systemctl 2>/dev/null | grep '\\-\\.mount' " ..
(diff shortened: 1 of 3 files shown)


cartridge/failover.lua
```diff
@@ -200,6 +200,33 @@ local function apply_config(mod)
     )
 end
 
+--- Wait for box.info.vclock[id] to reach desired LSN.
+--
+-- @function wait_vclock
```
rosik, line 205: wait_lsn

cartridge/utils.lua
```diff
@@ -305,6 +306,33 @@ local function http_read_body(req)
...
+-- @tparam number lsn
+-- @tparam number pause
+-- @tparam number timeout
+-- @treturn boolean true / false
+local function wait_lsn(id, lsn, pause, timeout)
+    checks('number', 'number', 'number', 'number')
+    local deadline = fiber.time() + timeout
+
+    while true do
+        if (box.info.vclock[id] or 0) >= lsn then
```
olegrok, line 324:
```
0th vclock component will be used to count replica-local rows of an
anonymous replica. These rows won't be replicated and different
instances will have different values in vclock[0].
```

(https://github.com/tarantool/tarantool/commit/7106a492ccd75c9a5c1d68f886b71ca2a6835853)

I think we should add an assert if passed id is 0. Otherwise we don't give any guarantees about result.

cartridge/utils.lua
```diff
@@ -305,6 +306,33 @@ local function http_read_body(req)
...
+-- @local
+--
+-- @tparam number id
+-- @tparam number lsn
+-- @tparam number pause
+-- @tparam number timeout
+-- @treturn boolean true / false
+local function wait_lsn(id, lsn, pause, timeout)
+    checks('number', 'number', 'number', 'number')
+    local deadline = fiber.time() + timeout
```
olegrok, line 321: I suppose usage of `require('clock').monotonic()` is better when you work with timeouts.

cartridge/utils.lua
rosik: Should we also assert `id < 0` and `id > 32`?

olegrok: Not 32 but `box.schema.REPLICA_MAX`

Seems it should be `assert(0 < id and id < box.schema.REPLICA_MAX)`

cartridge/utils.lua
rosik: I've created new follow-up issue: https://github.com/tarantool/cartridge/issues/957

`fiber.time` is common for cartridge, so I'd rather resolve it aside.

cartridge/utils.lua
rosik: I would ignore these checks, otherwise it would require to check fractional values too, and it's a lot not very useful code.

PR #937, opened 2020-07-07
Fix another race in eventual failover

rosik:
Fix another race in eventual failover

Polling membership notifications in failover module is implemented using `fiber.cond:wait()`, which may miss an event if it arrives during some other yield. And it does in the following case:

1. Both leader and replica are healthy
2. Network blinks, and replica receives an event that leader is dead.
3. Replica starts roles reconfiguration, which may take some time.
4. Another rumor arrives, saying that leader is healthy, but replica misses it.
5. Replica finishes configuring roles, but the event is lost, and it stays rw.

This patch introduces **dispatching reconfiguration tasks is separate fiber**. Here is how it works:

1. Failover loop dispatches a task, which is initialized with `is_busy = nil` flag.
2. If the task meets all requirements (e.g. `wish_state()`) it sets `is_busy = true` and changes instance state to `ConfiguringRoles`.
3. Whenever another task arrives, all pending (not busy yet) tasks are cancelled. Busy tasks remain running and the new one gets scheduled.
4. Triggering two-phase commit similarly drops all tasks. It also expects that there are no busy tasks exist (their existence contradicts state machine restrictions).

I didn't forget about

- [x] Tests
- [x] Changelog


cartridge/failover.lua
```diff
@@ -43,7 +43,17 @@ local ApplyConfigError = errors.new_class('ApplyConfigError')
 local ValidateConfigError = errors.new_class('ValidateConfigError')
 local StateProviderError = errors.new_class('StateProviderError')
 
-vars:new('membership_notification', membership.subscribe())
+vars:new('membership_notification', fiber.channel(1))
+fiber.new(function()
```
olegrok, line 47: I suggest you to save this fiber into vars. I think it's not fine to create dangling fiber.

rosik:
I've reworked my patch. Previous version didn't take into account receiving new events during `wish_state()`.

cartridge/failover.lua
```diff
@@ -192,6 +200,49 @@ local function apply_config(mod)
...
 end
 
+local function reconfigure_all()
+    local confapplier = require('cartridge.confapplier')
+    local all_roles = require('cartridge.roles').get_all_roles()
+
+    -- WARNING: implicit yield
+    -- The event may arrive while two-phase commit is in progress.
+    -- We should wait for the appropriate state.
+    local state = confapplier.wish_state('RolesConfigured', math.huge)
```
olegrok, line 210: Why `math.huge`? Not some configurable value? e.g. two-phase commit timeout.

cartridge/failover.lua
```diff
@@ -192,6 +200,49 @@ local function apply_config(mod)
...
+    fiber.testcancel()
+
+    if state ~= 'RolesConfigured' then
+        log.info('Skipping failover step - state is %s', state)
+        return
+    end
+
+    fiber.self().storage.is_busy = true
```
olegrok, line 218: May be "configuration_is_in_progress" is better

cartridge/failover.lua
```diff
@@ -192,6 +200,49 @@ local function apply_config(mod)
...
+
+        return true
+    end)
+
+    if ok then
+        log.info('Failover step finished')
+    else
+        log.warn('Failover step failed: %s', err)
+    end
+    confapplier.set_state('RolesConfigured')
```
olegrok, line 243: Seems we should turn off `is_busy` mode here

cartridge/failover.lua
rosik: Because the only reason to abort reconfiguration is fiber cancellation and inappropriate state.
Bot of them are handled in this expression: wish_state returns immediately if state transition is impossible (e.g. OperationError -> RolesConfigured)

cartridge/failover.lua
rosik: It's not necessary. When the fiber is dead we can't access its storage anyway. So I check `fiber.status() == 'dead'` instead

cartridge/failover.lua
rosik: I would rather not. It doesn't add much new information.

cartridge/failover.lua
```diff
@@ -200,65 +251,53 @@ local function failover_loop(args)
...
-        if ok then
-            log.info('Failover step finished')
-        else
-            log.warn('Failover step failed: %s', err)
+        -- Schedule new task
+        do
+            local task = fiber.new(reconfigure_all, vars.cache.active_leaders)
+            local id = task:id()
+            task:name('cartridge.failover.task')
+            vars.schedule[id] = task
```
olegrok, line 294: Is order important here?

I assume that only last task should win, but seems it strictly depends on tarantool fiber scheduler.

rosik: There is always at most 1 busy task, and at most 1 pending.
Scheduling doesn't affect execution priority.

rosik: Order isn't important.
The only reason to use `schedule[id] = task` is that it looks prettier than `schedule[task] = true` (IMO).

PR #945, opened 2020-07-14
Implement vclock keeper storage in stateboard

### LINKED ISSUE #905: [4pt] Implement vclock-keepers storge in stateboard
Prerequisite for #749

rosik:
Vclockkeeper is a new concept required for consistent switchover. Vclockkeeper is an instance which is the most recent leader in replicaset. Every instance should persist his vclock in external storage before becoming rw. The update is protected by compare and swap (CaS) logic.

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)
- [x] Documentation (debt #958)

Close #905

### DIFF
--- a/cartridge/stateboard-client.lua
+++ b/cartridge/stateboard-client.lua
@@ -90,6 +90,69 @@ local function get_coordinator(session)
     return coodinator
 end
 
+local function set_vclockkeeper(session, replicaset_uuid, instance_uuid, vclock)
+    checks('stateboard_session', 'string', 'string', '?table')
+    assert(session.connection ~= nil)
+
+    local keeper, err = errors.netbox_call(session.connection,
+        'get_vclockkeeper', {replicaset_uuid},
+        {timeout = session.call_timeout}
+    )
+    if err ~= nil then
+        return nil, SessionError:new(err)
+    end
+
+    local ordinal = nil
+    if keeper == nil then
+        -- No keeper persisted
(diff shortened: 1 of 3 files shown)


cartridge/stateboard-client.lua
```diff
@@ -90,6 +90,69 @@ local function get_coordinator(session)
     return coodinator
 end
 
+local function set_vclockkeeper(session, replicaset_uuid, instance_uuid, vclock)
+    checks('stateboard_session', 'string', 'string', '?table')
+    assert(session.connection ~= nil)
+
+    local keeper, err = errors.netbox_call(session.connection,
```
olegrok, line 97: I don't understand a bit this part. Why do you call "get" and then "set" (two network calls). I assume you could do it via single call. Even it will be called "compare_and_set".

rosik: Well, I could move it outside the client, but it would save us only one `get_vclockkeeper` call in `failover.lua`. The coordinator role still has to query keepers before setting them because it _must_ be protected with CaS. We don't want any late TCP packet to void consistency.

I can try to optimize it later, but it would just increase quantity of code.

olegrok: > We don't want any late TCP packet to void consistency.

If I correctly understand you it could be eliminated via introducing critical section here.
Anyway if you want to stay it as is please file a follow-up issue to don't forget to optimize it in future.

PR #951, opened 2020-07-16
Test late consistent switchover after 2pc

rosik:


test/integration/switchover_test.lua
```diff
@@ -0,0 +1,129 @@
+local log = require('log')
```
Steap2448, line 1: `log` is unused

PR #959, opened 2020-07-22
Enhance logging in tests

rosik:
Log every init and apply_config within permanent role

I didn't forget about

- [x] Tests (unnecessary)
- [x] Changelog (unnecessary)

### DIFF
--- a/test/entrypoint/srv_basic.lua
+++ b/test/entrypoint/srv_basic.lua
@@ -102,11 +102,19 @@ package.preload['mymodule-dependency'] = function()
 end
 
 package.preload['mymodule-permanent'] = function()
+    local log = require('log')
     return {
         role_name = 'myrole-permanent',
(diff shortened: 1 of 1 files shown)


test/entrypoint/srv_basic.lua
```diff
@@ -102,13 +102,21 @@ package.preload['mymodule-dependency'] = function()
...
+    local log = require('log')
     return {
         role_name = 'myrole-permanent',
         permanent = true,
         get_role_name = function()
             return 'myrole-permanent'
         end,
 
+        init = function(opts)
+            log.info('--- init({is_master = %s})', opts.is_master)
```
olegrok, line 114: you could also add instance uuid or uri here

rosik: They're already logged by luatest

PR #962, opened 2020-07-23
Fix cfg_errors test compatibility with 2.3

rosik:
Error handling of `console.listen()` was changed recently, so we update
the test appropriately.

See: https://github.com/tarantool/tarantool/issues/4634

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)

### DIFF
--- a/test/unit/cfg_errors_test.lua
+++ b/test/unit/cfg_errors_test.lua
@@ -238,18 +238,29 @@ g.test_console_sock = function()
     g.mock_membership()
 
     local sock_dir = fio.pathjoin(g.tempdir, 'sock')
+    local sock_name = string.format('%s/%s.sock', sock_dir, ('a'):rep(110))
     fio.mktree(sock_dir)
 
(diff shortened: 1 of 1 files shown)


test/unit/cfg_errors_test.lua
```diff
@@ -238,18 +238,26 @@ g.test_console_sock = function()
...
+        advertise_uri = 'unused:0',
+        http_enabled = false,
+        roles = {},
+        console_sock = sock_name,
+    })
+
+    t.assert_not(ok)
+    t.assert_covers(err, {class_name = 'CartridgeCfgError'})
     t.assert_equals(fio.listdir(sock_dir), {})
+    local e1 = 'Too long console_sock exceeds UNIX_PATH_MAX limit'
```
olegrok, line 255: I suggest you to add some comments that one line is needed for "before Tarantool 2.3" and another for "Tarantool 2.3+"

rosik: Done

PR #963, opened 2020-07-23
Fix feedback test compatibility with 1.10

rosik:
The feedback is unable to find manifest under tarantool 1.10 because it
doesn't provide `package.searchroot()`. This patch gracefully skips it.

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)

### DIFF
--- a/test/integration/feedback_test.lua
+++ b/test/integration/feedback_test.lua
@@ -50,6 +50,10 @@ g.before_all = function()
         },
     })
     g.cluster:start()
+
+    g.msg = ch:get(2)
+    t.assert_type(g.msg, 'table', 'No feedback received')
+    log.info('Feedback: %s', json.encode(g.msg))
 end
 
 g.after_all = function()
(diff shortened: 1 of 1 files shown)


test/integration/feedback_test.lua
```diff
@@ -50,6 +50,10 @@ g.before_all = function()
         },
     })
     g.cluster:start()
+
+    g.msg = ch:get(2)
+    t.assert_type(g.msg, 'table', 'No feedback received')
```
olegrok, line 55: Could you check msg content here?

rosik: assert_type already prints value if type mismatches

```console
tarantool> t = require('luatest')
---
...

tarantool> t.assert_type('foo', 'table', 'something serious is broken')
---
- error:
    status: fail
    class: LuatestError
    message: 'something serious is broken

      expected: a table value, actual: type string, value "foo"'
...
```

And other fields are checked in tests.

PR #965, opened 2020-07-23
Add timeout for membership probe_uri

vanyarock01:

I didn't forget about

- [x] Tests
- [x] Changelog
- [ ] Documentation

Close #???

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,7 @@ Added
   failure, a user could either revert it (promote previous leader), or force
   promotion to be inconsistent.
 - Early logger initialization (for Tarantool > 2.5.0-100, which supports it).
+- Probe uri timeout in cartridge.cfg.
 
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Fixed
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,7 +22,8 @@ Added
   failure, a user could either revert it (promote previous leader), or force
   promotion to be inconsistent.
 - Early logger initialization (for Tarantool > 2.5.0-100, which supports it).
-- Probe uri timeout in cartridge.cfg.
+- Add `probe_uri_timeout` argparse option responsible for retrying
+  "Can't ping myself" error on startup.
 
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Fixed
--- a/cartridge.lua
+++ b/cartridge.lua
@@ -19,6 +19,7 @@ local errors = require('errors')
 local membership = require('membership')
 local membership_network = require('membership.network')
 local http = require('http.server')
+local fiber = require('fiber')
 
 local rpc = require('cartridge.rpc')
 local auth = require('cartridge.auth')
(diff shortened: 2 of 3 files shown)


cartridge.lua
```diff
@@ -216,6 +217,8 @@ local function cfg(opts, box_opts)
         console_sock = '?string',
         webui_blacklist = '?table',
         upgrade_schema = '?boolean',
+        probe_uri_timeout = '?number',
+        probe_uri_delay = '?number',
```
rosik, line 221: Options variety increase complexity. I would recommend acquiring it with `argparse` and keeping it private.

cartridge.lua
```diff
@@ -461,9 +464,23 @@ local function cfg(opts, box_opts)
...
+        if ok then
+            resolved = true
+            break
+        end
+        log.info('Can not ping myself: %s', estr)
+        time = time - delay
+    end
+
+    if not resolved then
+        return nil, CartridgeCfgError:new('Instance has not been resolved')
```
rosik, line 483: This loop is incorrect and too verbose.

```
while true do 
  local ok, err = ...
  if ok then
    break
  elseif time > deadline then
    return nil, err
  else 
    fiber.sleep(pause)
  end
end
```

cartridge.lua
```diff
@@ -461,9 +464,23 @@ local function cfg(opts, box_opts)
...
+    local resolved = false
+
+    while time > 0 do
+        local ok, estr = membership.probe_uri(membership.myself().uri)
+        if ok then
+            resolved = true
+            break
+        end
+        log.info('Can not ping myself: %s', estr)
+        time = time - delay
```
rosik, line 479: What is the purpose of delay?

vanyarock01: This is the delay between checks

test/unit/cfg_errors_test.lua
```diff
@@ -263,6 +278,71 @@ g.test_console_sock = function()
...
+
+    -- default probe_uri_timeout: 1.0s
+    local ok, err = cartridge.cfg({
+        workdir = g.tempdir,
+        advertise_uri = 'localhost:13001',
+        roles = {},
+    })
+    t.assert(ok, err)
+
+    stop_http_service()
```
rosik, line 316: better use `http_enabled = false`

test/unit/cfg_errors_test.lua
```diff
@@ -263,6 +278,71 @@ g.test_console_sock = function()
...
+        fiber.sleep(0.3)
+        dns_resolved = true
+    end)
+
+    -- default probe_uri_timeout: 1.0s
+    local ok, err = cartridge.cfg({
+        workdir = g.tempdir,
+        advertise_uri = 'localhost:13001',
+        roles = {},
+    })
```
rosik, line 313: You were going to make default 0. I suggest passing timeout explicitly with `os.setenv`

cartridge.lua
```diff
@@ -461,9 +462,27 @@ local function cfg(opts, box_opts)
...
+    end
+
+    local delay = require('membership.options').ACK_TIMEOUT_SECONDS
+    local deadline = fiber.time() + (probe_uri_opts.probe_uri_timeout or 0)
+
+    while true do
+        local next_wakeup = fiber.time() + delay 
+        local ok, estr = membership.probe_uri(membership.myself().uri)
+        local now = fiber.time()
+        if ok then
```
vasiliy-t, line 478: consider adding log.verbose("it's ok .....") could be useful in case of debug

cartridge.lua
```diff
@@ -461,9 +462,27 @@ local function cfg(opts, box_opts)
...
+            break
+        elseif now >= deadline then
+            return nil, CartridgeCfgError:new('Can not ping myself: %s', estr)
+        else
+            log.info('Can not ping myself: %s', estr)
```
vasiliy-t, line 483: consider logging remaining time

cartridge.lua
```diff
@@ -461,9 +462,27 @@ local function cfg(opts, box_opts)
...
+    local deadline = fiber.time() + (probe_uri_opts.probe_uri_timeout or 0)
+
+    while true do
+        local next_wakeup = fiber.time() + delay 
+        local ok, estr = membership.probe_uri(membership.myself().uri)
+        local now = fiber.time()
+        if ok then
+            break
+        elseif now >= deadline then
+            return nil, CartridgeCfgError:new('Can not ping myself: %s', estr)
```
vasiliy-t, line 481: "can not ping myself" because of what? consider adding something like "deadline exceeded" to log message

rosik: because of `%s, estr`

CHANGELOG.rst
```diff
@@ -22,6 +22,7 @@ Added
   failure, a user could either revert it (promote previous leader), or force
   promotion to be inconsistent.
 - Early logger initialization (for Tarantool > 2.5.0-100, which supports it).
+- Probe uri timeout in cartridge.cfg.
```
rosik, line 25:
```suggestion
- Add `probe_uri_timeout` argparse option responsible for retrying
  "Can't ping myself" error on startup.
```

PR #966, opened 2020-07-23
Parametrize stateful failover tests

rosik:
Run the same tests for both stateboard and etcd2.
Reshuffle test code to make it compatible with both cases.
Also fix some flaky bugs:
- call `wish_state` before `apply_config`;
- add few more checks for the sake of synchronization;
- fix mistake in promotion error message;

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)

### DIFF
--- a/test/integration/failover_stateful_test.lua
+++ b/test/integration/failover_stateful_test.lua
@@ -1,41 +1,24 @@
 local log = require('log')
 local fio = require('fio')
 local t = require('luatest')
-local g = t.group()
-
+local httpc = require('http.client')
 local helpers = require('test.helper')
 
+local etcd2_client = require('cartridge.etcd2-client')
+local stateboard_client = require('cartridge.stateboard-client')
+
+local g_etcd2 = t.group('integration.failover_stateful.etcd2')
+local g_stateboard = t.group('integration.failover_stateful.stateboard')
+
 local storage_uuid = helpers.uuid('b')
-local storage_1_uuid = helpers.uuid('b', 'b', 1)
-local storage_2_uuid = helpers.uuid('b', 'b', 2)
-local storage_3_uuid = helpers.uuid('b', 'b', 3)
+local S1; local storage_1_uuid = helpers.uuid('b', 'b', 1)
+local S2; local storage_2_uuid = helpers.uuid('b', 'b', 2)
+local S3; local storage_3_uuid = helpers.uuid('b', 'b', 3)
 
 local router_uuid = helpers.uuid('a')
-local router_1_uuid = helpers.uuid('a', 'a', 1)
+local R1; local router_1_uuid = helpers.uuid('a', 'a', 1)
 
-g.before_all(function()
-    g.datadir = fio.tempdir()
-
-    fio.mktree(fio.pathjoin(g.datadir, 'stateboard'))
-    g.kvpassword = require('digest').urandom(6):hex()
-    g.stateboard = require('luatest.server'):new({
-        command = helpers.entrypoint('srv_stateboard'),
-        workdir = fio.pathjoin(g.datadir, 'stateboard'),
-        net_box_port = 14401,
-        net_box_credentials = {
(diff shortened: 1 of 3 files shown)


olegrok:
> Also fix some bugs

Please list them in commit message

test/integration/failover_stateful_test.lua
```diff
@@ -66,36 +50,155 @@ g.before_all(function()
...
+            '--data-dir', self.workdir,
+            '--listen-peer-urls', 'http://localhost:17001',
+            '--listen-client-urls', URI,
+            '--advertise-client-urls', URI,
+        })
+
+        helpers.retrying({}, function()
+            local resp = httpc.put(URI .. '/v2/keys/hello?value=world')
+            assert(resp.status == 200, resp.body)
+            require('log').info(
```
olegrok, line 130: require is redundant

test/integration/failover_stateful_test.lua
```diff
@@ -66,36 +50,155 @@ g.before_all(function()
...
+        end)
+    end
+    function g.state_provider:stop()
+        local process = self.process
+        if process == nil then
+            return
+        end
+        self.process:kill()
+        helpers.retrying({}, function()
+            t.assert_not(process:is_alive(), 'Etcd is still running')
```
olegrok, line 142: IMO you shouldn't fail after timeout. Just send SIGKILL if it's still alive.

test/integration/failover_stateful_test.lua
```diff
@@ -66,36 +50,155 @@ g.before_all(function()
...
+        require('log').info('Coordinator options updated')
     ]])
-    g.cluster.main_server.net_box:call(
+end
+
+g_stateboard.before_all(function()
+    local g = g_stateboard
+    g.datadir = fio.tempdir()
+
+    fio.mktree(fio.pathjoin(g.datadir, 'stateboard'))
```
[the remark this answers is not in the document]
rosik: It's redundant. Removed it completely.

test/integration/failover_stateful_test.lua
rosik: Fixed

test/integration/failover_stateful_test.lua
rosik: I don't want to:

1. It's copy-pasted from test helpers
2. There's no convenient way to do it.

test/integration/failover_stateful_test.lua
```diff
@@ -132,107 +246,103 @@ function g.test_stateboard_restart()
         }})
     end)
 
-    g.stateboard:start()
+    fio.rmtree(g.state_provider.workdir)
```
olegrok, line 249: Do you really need it? If you drop state_provider data you don't restart it. You boot completely new state_provider instance.

test/integration/failover_stateful_test.lua
```diff
@@ -132,107 +246,103 @@ function g.test_stateboard_restart()
...
-        }
-    )
+end)
 
+add('test_leader_restart', function(g)
     -----------------------------------------------------
-    g.cluster:server('storage-1'):stop()
-    t.assert_equals(
-        g.stateboard.net_box:call('longpoll', {3}),
+    g.client:longpoll(0)
```
olegrok, line 313: Why did you add longpoll(0) here?

test/integration/failover_stateful_test.lua
rosik: Yes.

1. The intention was to cover the case of clean state provider, as you correctly pointed out.
2. This patch doesn't add anything new. This line originally was above `:stop()`, but sometimes it produces unnecessary warnings in logs, so I moved it here.

test/integration/failover_stateful_test.lua
rosik: It's already been there. It's purpose is to abandon all previous events and act as a sync point of S1 death.

PR #973, opened 2020-07-28
Fix flaky state_machine test

rosik:
In test_fiber_cancel we check that instance becomes a leader under some
circumstances, but the check is incorrect. It queries if instance is writable
in several consequent calls, but the instance may trigger state more quickly.

With this patch the check is performed using async call, which tells if
an instance was writable at least for a moment.

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)


olegrok:
Typo in commit message: `more quick` -> `more quickly`

rosik:
> Typo in commit message: `more quick` -> `more quickly`

Ok, I'll fix it when squashing

PR #986, opened 2020-07-31
Avoid FD inheritance on forking

rosik:
Workaround for https://github.com/tarantool/tarantool/issues/5220. If
tarantool uses logging into pipe, remote-control fd is inherited by the
forked process and never closed. It makes further box.cfg listen to fail
because address already in use.

Tarantool doesn't provide any legal way to set FD_CLOEXEC flag, so we do
it with ffi

I didn't forget about

- [x] Tests
- [x] Changelog

### DIFF
--- a/cartridge/utils.lua
+++ b/cartridge/utils.lua
@@ -1,10 +1,12 @@
 local fio = require('fio')
 local ffi = require('ffi')
+local bit = require('bit')
 local errno = require('errno')
 local fiber = require('fiber')
 local checks = require('checks')
 local errors = require('errors')
 
+local FcntlError = errors.new_class('FcntlError')
 local OpenFileError = errors.new_class('OpenFileError')
 local ReadFileError = errors.new_class('ReadFileError')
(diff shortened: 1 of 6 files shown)


cartridge/utils.lua
```diff
@@ -333,6 +339,41 @@ local function wait_lsn(id, lsn, pause, timeout)
...
+    checks('number')
+
+    local ret = ffi.C.fcntl(fd, F_GETFD)
+    if ret < 0 then
+        return nil, FcntlError:new(
+            "fcntl(F_GETFD) failed: %s", errno.strerror()
+        )
+    end
+
+    local flags = ffi.cast('uint64_t', bit.bor(ret, FD_CLOEXEC))
```
olegrok, line 365: I think you could check if such flag is already set. And if `bit.band(ret, FD_CLOEXEC) ~= 0` you can return true immediately.

rosik: Agree, fixed.

PR #989, opened 2020-08-06
Invalidate user's cookie after password change

### LINKED ISSUE #72: Invalidate session after password change
In GitLab by @rosik on Mar 22, 2019, 03:56

In context of #68

### LINKED ISSUE #840: Consider a way to invalidate cookie
Seems that if "lsid" is stolen administrator/user won't have a way to invalidate it.
Also I expect that lsid will be invalidated after password changing. But as I see "lsid" is saved after that.

It could be considered as security vulnerability as such cases.

olegrok:
This patch introduces new field in user model - version. We will
increment version after each password change. This approach allows
us to deny all sessions that were created before password was
changed.

Closes #840
Closes #72

### DIFF
--- a/cartridge/auth-backend.lua
+++ b/cartridge/auth-backend.lua
@@ -119,7 +119,8 @@ local function add_user(username, password, fullname, email)
         fullname = fullname,
         email = email,
         created_at = fiber.time64(),
-        password_data = create_password(password)
+        password_data = create_password(password),
+        version = 1,
     }
 
     local ok, err = cartridge.config_patch_clusterwide({users_acl = users_acl})
@@ -160,6 +161,11 @@ local function edit_user(username, password, fullname, email)
 
     if password ~= nil then
         user.password_data = update_password(password, user.password_data)
+        if user.version == nil then
+            user.version = 1
+        else
+            user.version = user.version + 1
+        end
     end
 
     if uid == nil then
--- a/cartridge/auth.lua
+++ b/cartridge/auth.lua
@@ -290,6 +290,25 @@ local function get_basic_auth_uid(auth)
     end
 end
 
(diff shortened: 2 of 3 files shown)


cartridge/auth.lua
```diff
@@ -377,6 +386,7 @@ local function coerce_user(user)
         username = user.username,
         fullname = user.fullname,
         email = user.email,
+        version = user.version,
```
rosik, line 389: Update the ldoc above too.

cartridge/auth-backend.lua
```diff
@@ -160,6 +161,11 @@ local function edit_user(username, password, fullname, email)
 
     if password ~= nil then
         user.password_data = update_password(password, user.password_data)
+        if user.version == nil then
+            user.version = 1
+        else
+            user.version = user.version + 1
+        end
```
rosik, line 168: It would be nice to preserve the session which updates the password. Otherwise the next request will redirect to authorization page

olegrok: I start to refresh current user session. Add new test case.

cartridge/auth.lua
```diff
@@ -179,14 +179,15 @@ local function get_params()
...
-    checks('string')
+local function create_cookie(uid, version)
+    checks('string', '?number')
     local ts = tostring(fiber.time())
     local key = cluster_cookie.cookie()
 
     local cookie = {
         ts = ts,
         uid = uid,
+        version = version,
```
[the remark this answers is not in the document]
olegrok: Yes, thanks, fixed

cartridge/auth.lua
olegrok: Fixed

Issue #908, opened 2020-06-19
Implement consistent promotion WebUI


rosik:
Follow-up for #749
Blocked by #907

rosik:
Resolved in #1001, #1010

PR #1000, opened 2020-08-11
Indicate consistent switchover problems in WebUI

rosik:

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)
- [x] Documentation

Part of #908

### DIFF
--- a/cartridge/lua-api/failover.lua
+++ b/cartridge/lua-api/failover.lua
@@ -144,6 +144,7 @@ end
 -- @tparam table { [replicaset_uuid] = leader_uuid }
 -- @tparam[opt] table opts
 -- @tparam ?boolean opts.force_inconsistency
+--   (default: **false**)
 --
 -- @treturn[1] boolean true On success
 -- @treturn[2] nil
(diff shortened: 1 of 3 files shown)


cartridge/lua-api/failover.lua
```diff
@@ -170,27 +171,30 @@ local function promote(replicaset_leaders, opts)
         return nil, PromoteLeaderError:new('There is no active coordinator')
     end
 
-    local _, err = rpc.call(
+    local ok, err = rpc.call(
```
olegrok, line 174: Don't understand why this change is necessary

cartridge/lua-api/failover.lua
```diff
@@ -170,27 +171,30 @@ local function promote(replicaset_leaders, opts)
...
     end
 
-    if opts == nil or opts.force_inconsistency == false then
-        return true
+    if opts ~= nil and opts.force_inconsistency == true then
+        local ok, err = failover.force_inconsistency(replicaset_leaders)
+        if ok == nil then
+            return nil, PromoteLeaderError:new(
+                "Promotion succeeded, but inconsistency wasn't forced: %s",
+                errors.is_error_object(err) and err.err or err
```
olegrok, line 190: Why isn't it simply `tostring(err)`. Seems it should work in the same way for simple strings and errors from "errors" module

rosik: tostring(err) includes stack traceback. Since I create new error object, i don't want it to become a part of new error message.

cartridge/lua-api/failover.lua
rosik: It isn't necessary. I can remove it if you insist.

olegrok: It depends on the way how errors are handled in cartridge. I'd prefer to use `err ~= nil` but if `ok == nil` is preferable for the cartridge, then you should leave this as is.

rosik: I think checking `ret == nil` is more suitable in general. The only exception are methods that can return `nil` without an error, but it's not the case

rosik: So I'll leave it as is.

PR #1004, opened 2020-08-12
Enhance vclockkeepers logging in stateboard

rosik:
It turns out `log.info()` doesn't yield. Thus the ugly hack with logging in a separate fiber isn't necessary.
Also vclockkeepers updates are logged.

Follow-up for #945

### DIFF
--- a/cartridge/stateboard.lua
+++ b/cartridge/stateboard.lua
@@ -84,6 +84,13 @@ local function set_leaders_impl(leaders)
         })
     end
     notification:broadcast()
+
+    box.on_commit(function()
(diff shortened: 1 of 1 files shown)


cartridge/stateboard.lua
```diff
@@ -159,6 +159,13 @@ local function set_vclockkeeper_impl(replicaset_uuid, instance_uuid, ordinal, vc
         }
     )
 
+    box.on_commit(function()
+        fiber.new(log.info, 'New vclockkeeper %s -> %s%s',
```
olegrok, line 163: Could be done without fiber creation

Issue #950, opened 2020-07-16
[2pt] Implement stateful failover fencing


rosik:
When external storage becomes unreachable enter read-only mode.

rosik:
# RFC

Neither eventual nor stateful failover modes don't protect a replicaset from the presence of multiple leaders when the network is partitioned. But fencing does. It enforces at-most-one leader policy in a replicaset.

Fencing operates as a fiber that occasionally checks connectivity with the state provider and with replicas. Fencing fiber runs on vclockkeepers; it starts right after consistent promotion succeeds. Replicasets which don't need consistency (single-instance and `all_rw`) don't defense, though. 

The condition for fencing actuation is the loss of both the state provider quorum and at least one replica. Otherwise, if either state provider is healthy or all replicas are alive, the fencing fiber waits and doesn't intervene.

When fencing is actuated, it generates a fake appointment locally and sets the leader to `nil`. Consequently, the instance becomes read-only. Subsequent recovery is only possible when the quorum reestablishes; replica connection isn't a must for recovery. Recovery is performed according to the rules of consistent switchover unless some other instance has already been promoted to a new leader.


Fencing is configured with the following options:

* ``FENCING_ENABLED``;

* ``FENCING_PAUSE`` -- the period of performing the check;

* ``FENCING_TIMEOUT`` -- time to actuate fencing after the check fails;

* ``FAILOVER_TIMEOUT`` -- time to declare an unresponsive member as dead
  (not a fencing option actually, it'll affect other failover modes too).

All of them will be a part of `failover_get/set_params` API.

rosik:
Вопросы от коллег:

Возможно стоит объединить параметры таймаутов,
Дать рекомендации по выбору значений.
Посмотреть в другие базы - как у них сделано
Посмотреть блог афира - кто лучше переживает сплитбрейн

rosik:
Я уверен, что fencing_timeout и failvoer_timeout не стоит объединять. Сделать одну из них конфигурироемой неявно можно, но явное лучше неявного.

Что касается других баз - 

Redis sentinel: down-after-milliseconds, failover-timeout (https://redis.io/topics/sentinel)
Cocroachdb: time_until_store_dead (https://www.cockroachlabs.com/docs/stable/demo-fault-tolerance-and-recovery.html)

PR #1011, opened 2020-08-14
Split etcd into a separate test helper

rosik:
Starting etcd becomes popular in tests, so it's split into a separate test helper.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Part of #950

### DIFF
--- a/cartridge/test-helpers/etcd.lua
+++ b/cartridge/test-helpers/etcd.lua
@@ -64,11 +64,9 @@ end
 
 --- Start the node.
 function Etcd:start()
-    local env = os.environ()
     local log_cmd = {}
     for k, v in pairs(self.env) do
         table.insert(log_cmd, string.format('%s=%q', k, v))
-        env[k] = v
     end
     table.insert(log_cmd, self.command)
     for _, v in ipairs(self.args) do
@@ -77,7 +75,7 @@ function Etcd:start()
 
     log.debug(table.concat(log_cmd, ' '))
 
-    self.process = Process:start(self.etcd_path, self.args, env, {
+    self.process = Process:start(self.etcd_path, self.args, self.env, {
         output_prefix = 'etcd-' .. self.name,
     })
     log.debug('Started server PID: ' .. self.process.pid)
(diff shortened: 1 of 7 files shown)


cartridge/test-helpers/etcd.lua
```diff
@@ -0,0 +1,114 @@
...
+        ETCD_DATA_DIR = self.workdir,
+        ETCD_LISTEN_PEER_URLS = self.peer_url,
+        ETCD_LISTEN_CLIENT_URLS = self.client_url,
+        ETCD_ADVERTISE_CLIENT_URLS = self.client_url,
+    }, self.env):tomap()
+end
+
+--- Start the node.
+function Etcd:start()
+    local env = table.copy(os.environ())
```
olegrok, line 67:
```
tarantool> os.environ()['a'] = 2
---
...

tarantool> os.environ()['a']
---
- null
...
```

Seems table.copy is redundant

cartridge/test-helpers/etcd.lua
```diff
@@ -0,0 +1,114 @@
...
+        env[k] = v
+    end
+    table.insert(log_cmd, self.command)
+    for _, v in ipairs(self.args) do
+        table.insert(log_cmd, string.format('%q', v))
+    end
+
+    log.debug(table.concat(log_cmd, ' '))
+
+    self.process = Process:start(self.etcd_path, self.args, self.env, {
```
olegrok, line 80: `self.env` -> `env`?

cartridge/test-helpers/etcd.lua
rosik: You're right, it [returns](https://github.com/tarantool/tarantool/blob/0ab21ac9eeaaae2aa0aef5e598d374669f96df9e/src/lua/env.lua#L12-L28) new table every time. Fixed

cartridge/test-helpers/etcd.lua
rosik: Fixed

PR #1031, opened 2020-08-31
Use replication_connect_quorum = 0 by default

rosik:
In v2.0.0 the state machine was introduced. One of its features is a "ConnectingFullmesh" state. It was designed to prevent leader from becoming rw after restart unless replication sync succeeds. This is especially desirable with eventual failover mode.

But the same "ConnectingFullmesh" became a problem because instances stuck in case of replica unavailability. And the same problem occurs when `advertise_uri` is changed. As a solution we proposed to explicitly set quorum to 0.

This patch inverts default behavior - now instances wouldn't stuck, but if a user wants, he may set quorum > 0 explicitly.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation


rosik:
Closing in favor of #1032

PR #1032, opened 2020-08-31
Avoid ConnectingFullmesh trap on advertise_uri change

rosik:
On restart cartrige composes replication config based on clusterwide
configuration. Restarting an instance with a different `advertise_uri`
results in `box.cfg({replication = old_uri, quorum = 1})`.
And since the `old_uri` is unavailable, it gets stuck.

This patch changes the way of replication config generation: present
`advertise_uri` is used for the current instance instead of an old URI.

I didn't forget about

- [x] Tests
- [x] Changelog


test/integration/reboot_test.lua
```diff
@@ -198,3 +198,15 @@ function g.test_invalid_config()
...
+function g.test_advertise_change()
+    local master = g.cluster.main_server
+    master:stop()
+
+    master.env['TARANTOOL_ADVERTISE_URI'] = '127.0.0.1:13310'
+    master.net_box_uri = '127.0.0.1:13310'
+    master:start()
+
+    helpers.wish_state(master, 'RolesConfigured')
+
```
olegrok, line 211: Redundant line

test/integration/reboot_test.lua
```diff
@@ -198,3 +198,15 @@ function g.test_invalid_config()
...
+function g.test_advertise_change()
+    local master = g.cluster.main_server
+    master:stop()
+
+    master.env['TARANTOOL_ADVERTISE_URI'] = '127.0.0.1:13310'
+    master.net_box_uri = '127.0.0.1:13310'
+    master:start()
+
+    helpers.wish_state(master, 'RolesConfigured')
```
olegrok, line 210: It would be great to specify timeout here

test/integration/reboot_test.lua
```diff
@@ -198,3 +198,15 @@ function g.test_invalid_config()
         " not in clusterwide config, no idea what to do now"
     })
 end
+
+function g.test_advertise_change()
+    local master = g.cluster.main_server
```
olegrok, line 203: The test is seems to be OK. But I think you can add more complex case when master is stopped, leader is changed and some DML or even DDL actions are made.
After that master is restarted with different advertise_uri.

cartridge/confapplier.lua
```diff
@@ -381,18 +382,11 @@ local function boot_instance(clusterwide_config)
...
-        -- Due to the bug box_opts.replication_connect_quorum was ignored
-        -- and box.cfg used to hang
-        table.remove(
-            box_opts.replication,
-            utils.table_find(
-                box_opts.replication,
-                pool.format_uri(vars.advertise_uri)
-            )
+            topology_cfg, replicaset_uuid,
+            -- Workaround for https://github.com/tarantool/tarantool/issues/3760
```
olegrok, line 386: Seems patch has been already cherry-picked https://github.com/tarantool/tarantool/issues/3760

Not sure we should support some workarounds.

rosik: I've removed stale comment, but I'd like to preserve the workaround - it doesn't affect anything.

test/integration/reboot_test.lua
rosik: Fixed

test/integration/reboot_test.lua
rosik: Done

test/integration/reboot_test.lua
rosik: I've added a ddl operation. There is no other instances  to change leader. Rebootstrap is already tested in `integration.rejoin.test_rejoin_*`

olegrok: What will happen if several instances will change its advertise-uri?

rosik: Unless you've specified `replication_connect_quorum = 0` it'll stay in ConnectingFullmesh.

rosik: (It's intentional)

olegrok: Well, but seems it should be documented behaviour

olegrok (approved):

PR #1034, opened 2020-09-02
GC pool connections

rosik:
Netbox connections established by `pool.connect` are cached, and it prevented them from being garbage-collected. Cartridge doesn't usually connect to all instances unless it serves WebUI. Before the patch there were no way to cleanup the pool if it became redundant.

This patch introduces two changes:
* Netbox connections established by `pool.connect` can be garbage-collected now.
* But pool preserves a reference for a while. The time is controlled by the new `refkeeper_duration` option.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation


rosik:
Rejected after discussion in chat.

Issue #566, opened 2020-02-10
[2pt] Recover from OperationError


rosik:
Re-applying configuration should be a valid case.

PR #1045, opened 2020-09-14
New patch_clusterwide option to avoid skipping 2pc

rosik:
Don't skip two-phase commit even if the patch doesn't introduce any changes.

By default it's now applied for the http upload API and for GraphQL mutation `cluster { sections }`.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Part of #566

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,16 @@ and this project adheres to
 -------------------------------------------------------------------------------
 
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Added
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- New option in ``cartridge.config_patch_clusterwide``: ``{never_skip = true}``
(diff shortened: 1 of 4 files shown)


CHANGELOG.rst
```diff
@@ -12,6 +12,16 @@ and this project adheres to
 [Unreleased]
 -------------------------------------------------------------------------------
 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Added
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- New option in ``cartridge.config_patch_clusterwide``: ``{never_skip = true}``
```
olegrok, line 19: Not sure about naming. Why "never" if it work only for one time. I believe we should have "force" in option name

cartridge/webui/api-config.lua
```diff
@@ -126,7 +126,7 @@ local function upload_config(clusterwide_config)
         end
     end
 
-    return twophase.patch_clusterwide(patch)
+    return twophase.patch_clusterwide(patch, {never_skip = true})
```
olegrok, line 129: I'm not sure we should enable it anywhere by default

rosik:
Rejected after chat discussion

PR #1049, opened 2020-09-16
Split advertise change test into a separate group

rosik:
It'll be useful for the future cases development.

### DIFF
--- a/test/integration/advertise_change_test.lua
+++ b/test/integration/advertise_change_test.lua
@@ -1,0 +1,86 @@
+local fio = require('fio')
+local fun = require('fun')
+local t = require('luatest')
+local g = t.group()
+
+local helpers = require('test.helper')
+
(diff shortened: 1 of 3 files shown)


test/integration/advertise_change_test.lua
```diff
@@ -0,0 +1,86 @@
...
+
+local function move(srv, uri)
+    srv.env['TARANTOOL_ADVERTISE_URI'] = uri
+    srv.net_box_uri = uri
+end
+
+g.before_all(function()
+    g.cluster = helpers.Cluster:new({
+        datadir = fio.tempdir(),
+        use_vshard = false,
```
olegrok, line 16: Why not?

rosik: For faster tests. It's not related to this case

olegrok (approved):

PR #1050, opened 2020-09-16
Refine uri when querying topology

### LINKED ISSUE #1037: [2pt] Respect new advertise_uri when querying servers
Follow-up for #1033 

GraphQL query `{servers {}}` returns uri from clusterwide config now. Instead it should return proper uri (from the new `refine_servers_uri` function).

It also should be used for querying `box_info` and server `statistics`.

rosik:
The old algorithm was getting info from a member with the latest timestamp.
There are two problems with this approach:

1. Timestamp is declared as internal in membership module.
2. It's only updated when new gossip arrives.

In some cases it could result in status "blinking".

This patch makes use of `topology.refine_servers_uri` to resemble other cartridge modules.

I didn't forget about

- [x] Tests
- [x] Changelog

Close #1037

### DIFF
--- a/cartridge/lua-api/get-topology.lua
+++ b/cartridge/lua-api/get-topology.lua
@@ -126,29 +126,27 @@ local function get_topology()
     end
 
     local active_leaders = failover.get_active_leaders()
+    local refined_uri = topology.refine_servers_uri(topology_cfg)
 
     for _, instance_uuid, server in fun.filter(topology.not_expelled, topology_cfg.servers) do
-        local uri = server.uri
+        local uri = assert(refined_uri[instance_uuid])
         local member = members[uri]
-        local alias = nil
-        if member and member.payload then
-            alias = member.payload.alias
-        end
+        members[uri] = nil
 
         local srv = {
             uri = uri,
             uuid = instance_uuid,
-            alias = alias,
+            disabled = not topology.not_disabled(instance_uuid, server),
+            alias = nil,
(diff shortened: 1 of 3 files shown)


cartridge/lua-api/get-topology.lua
```diff
@@ -200,12 +114,80 @@ local function get_topology()
...
+    -- @tfield number priority
+    --   Leadership priority for automatic failover.
+    -- @tfield number clock_delta
+    --   Difference between remote clock and the current one (in
+    --   seconds), obtained from the membership module (SWIM protocol).
+    --   Positive values mean remote clock are ahead of local, and vice
+    --   versa.
+    -- @table ServerInfo
     for _, instance_uuid, server in fun.filter(topology.not_expelled, topology_cfg.servers) do
-        local srv = get_server_info(members, instance_uuid, server.uri)
```
olegrok, line 205: To be honest I'd prefer to see this logic in separate function.

It looks a bit strange to describe arguments of function part right inside function - usually if you want to describe that some piece of code does you write extended comment and it's pure description for people (without ldoc).

rosik: This function should be executed in a context (it modifies `members` table), that's why i merged them.

olegrok: Well, remove ldoc here. And rewrite comment for humans

olegrok (approved):

PR #1059, opened 2020-09-21
Fix CI for tarantool 2.4+

### LINKED ISSUE #1058: [1pt] Fix remote_control_test for tarantool 2.4+
Tarantool box.error API was significantly refactored in scope of https://github.com/tarantool/tarantool/issues/4398 and now cartridge tests fail:

```lua
local ok, err = pcall(conn.eval, conn,
    'error("ScriptError", 0)'
)
t.assert_not(ok)
t.assert_equals(type(err), 'cdata')
t.assert_equals(err.code, box.error.PROC_LUA) -- expected: 32, actual: nil
```

https://gitlab.com/tarantool/cartridge/-/jobs/746467482

rosik:
CI for Tarantool 2.6 [fails](https://gitlab.com/tarantool/cartridge/-/jobs/754399625) for two reasons:

1. Errors API was significantly refactored in scope of https://github.com/tarantool/tarantool/issues/4398.

2. Recently in [1] we've monkey-patched `fio.tempdir` implementation because it didn't respect `$TMPDIR` env variable. In [2] we've extended it with `base` agrument because etcd required another temp directory (`/dev/shm` tmpfs in Gitlab CI runner wasn't large enough).  Later, TMPDIR was supported in tarantool (see [3]), but it still doesn't support passing `base` as an argument.

This patch fixes the second problem only - it completely eliminates the use of built-in `fio.tempdir` implementation in favor of ours.
The first one will be fixed separately (maybe in tarantool)

[1] https://github.com/tarantool/cartridge/pull/613
[2] https://github.com/tarantool/cartridge/pull/966
[3] https://github.com/tarantool/tarantool/issues/4794

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)

Part of #1058

### DIFF
--- a/test/helper.lua
+++ b/test/helper.lua
@@ -6,18 +6,13 @@ local helpers = table.copy(require('cartridge.test-helpers'))
 
 helpers.project_root = fio.dirname(debug.sourcedir())
 
-local __fio_tempdir = fio.tempdir
 fio.tempdir = function(base)
     base = base or os.getenv('TMPDIR')
-    if base == nil or base == '/tmp' then
-        return __fio_tempdir()
-    else
-        local random = digest.urandom(9)
-        local suffix = digest.base64_encode(random, {urlsafe = true})
-        local path = fio.pathjoin(base, 'tmp.cartridge.' .. suffix)
-        fio.mktree(path)
-        return path
-    end
+    local random = digest.urandom(9)
+    local suffix = digest.base64_encode(random, {urlsafe = true})
+    local path = fio.pathjoin(base, 'tmp.cartridge.' .. suffix)
+    fio.mktree(path)
+    return path
 end
 
 function helpers.entrypoint(name)


test/unit/remote_control_test.lua
```diff
@@ -530,7 +530,6 @@ function g.test_call()
         )
         t.assert_not(ok)
         t.assert_equals(type(err), 'cdata')
-        t.assert_equals(err.code, box.error.PROC_LUA)
```
olegrok, line 533: This way you just remove a part of tests. IMO it's not a solution. I think it's appropriate to support test cases for all versions based on _TARANTOOL value.

Personally, I'm not sure we should avoid tarantool internals consideration. Especially in case when implement Tarantool-compatible interfaces.

olegrok (approved):

PR #1073, opened 2020-09-28
Option to disable UDP broadcast

### LINKED ISSUE #1057: [2pt] New option to turn off autodiscavery for new nodes via udp
Hello!
We ecnountered the same issue already 3 times when our new devops deployed Tarantool cluster with cartridge via cartridge ansbile in to the same subnet with same same cookie as existing Tarantool cluster.
This issue leads to bugs in existed and new Tarantool clusters.
We don't need autodiscavery feature and i think we should have option to turn off this feature to prevent situations like this in future.

rosik:
Introducing new `cartridge.cfg` option `swim_broadcast` (default true) to disable instances discovery on start.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1057

### DIFF
--- a/test/unit/cfg_errors_test.lua
+++ b/test/unit/cfg_errors_test.lua
@@ -320,12 +320,14 @@ end
 
 g.test_positive = function()
     g.mock_membership()
+    membership.broadcast = function() error('Forbidden', 0) end
     -- Test successful cartridge.cfg
 
     local opts = {
             workdir = '/tmp',
             advertise_uri = 'unused:0',
             http_enabled = false,
+            swim_broadcast = false,
         roles = {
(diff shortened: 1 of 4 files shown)


test/unit/cfg_errors_test.lua
```diff
@@ -320,12 +320,14 @@ end
 
 g.test_positive = function()
     g.mock_membership()
+    membership.broadcast = function() error('Forbidden', 0) end
```
olegrok, line 323: I think error message should contain "Membership broadcast is forbidden" but not just "Forbidden"

olegrok (approved):
Small nit, but in general LGTM

PR #1108, opened 2020-10-30
Evict failover schedule functions

rosik:
Slightly reorganize code to eliminate repeats

I didn't forget about

- [x] Tests (preserved green)
- [x] Changelog (unnecessary)


cartridge/failover.lua
```diff
@@ -422,26 +451,10 @@ local function failover_loop(args)
...
-        do
-            local task = fiber.new(reconfigure_all, vars.cache.active_leaders)
-            local id = task:id()
-            task:name('cartridge.failover.task')
-            vars.schedule[id] = task
-            log.info('Failover triggered, reapply scheduled (fiber %d)', id)
-        end
+        log.info(
+            'Failover triggered, reapply scheduled (fiber %d)',
+            schedule_add(reconfigure_all, vars.cache.active_leaders)
```
olegrok, line 456: I don't like this idea. log function argument performs some complex logic inside. Let's declare a local variable, save result into it and then log it.

cartridge/failover.lua
```diff
@@ -78,6 +78,35 @@ function _G.__cartridge_failover_wait_rw(timeout)
     return errors.pcall('WaitRwError', box.ctl.wait_rw, timeout)
 end
 
+--- Cancel all pending tasks.
```
olegrok, line 81: Is there some description what "task" is?

cartridge/failover.lua
rosik: Fixed

cartridge/failover.lua
rosik: I've added a comment and reorganized it once more

cartridge/failover.lua
```diff
@@ -422,26 +453,8 @@ local function failover_loop(args)
             goto continue
```
olegrok, line 453:
```
LuajitError: error loading module 'cartridge.failover' from file '/builds/tarantool/cartridge/.rocks/share/tarantool/cartridge/failover.lua':
	.../cartridge/.rocks/share/tarantool/cartridge/failover.lua:446: <goto continue> jumps into the scope of local 'id'
fatal error, exiting the event loop
```

PR #1109, opened 2020-10-30
Disable consistency for single-instance replicaset

rosik:
Single-instance replicasets has no need to care about consistency.
This patch is inspired by fencing, which shouldn't set read-only
all_rw instances. And so it shouldn't affect loners.

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)
- [x] Documentation

### DIFF
--- a/rst/topics/failover.rst
+++ b/rst/topics/failover.rst
@@ -176,6 +176,11 @@ options: to revert promotion (to re-promote the old leader) or to force it
 inconsistently (all kinds of ``failover_promote`` API has
 ``force_inconsistency`` flag).
 
+Consistent promotion doesn't work for replicasets with `all_rw` flag enabled
+and for single-instance replicasets. In these two cases an instance doesn't
+even try to query `vclockkeeper` and to perform `wait_lsn`. But the coordinator
+still appoints a new leader if the current one dies.
+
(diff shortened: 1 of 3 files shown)


rst/topics/failover.rst
```diff
@@ -176,6 +176,11 @@ options: to revert promotion (to re-promote the old leader) or to force it
 inconsistently (all kinds of ``failover_promote`` API has
 ``force_inconsistency`` flag).
 
+Consistent promotion doesn't work for replicasets with `all_rw` flag enabled
+and for single-instance replicasets. In these two cases an instance doesn't
+even try to query `vclockkeeper` and to perform `wait_lsn`. But leader
+appointment rules still takes place -- if a leader dies, it's get switched.
```
olegrok, line 182: `it's get switched`

Is it correct? May be `it gets switched` or `it's gotten switched`?

olegrok (approved):

rst/topics/failover.rst
pgulutzan: Yes, "it gets switched" is much better grammar -- but it means "the leader gets switched", which is odd because the leader is dead. Perhaps the intent was to say "But leader appointment rules still apply -- if a leader dies, then a new leader is appointed."

rosik: Thanks. I've already changed it to "But the coordinator still appoints a new leader if the current one dies". Is it ok?

pgulutzan: That is better than what I suggested.

PR #1110, opened 2020-10-31
CI: add CI tests for MacOS

olegrok:
Enable MacOS tests with Gitlab Actions

Part of #1093
Close #1153 
Close #1069

PR #1115, opened 2020-11-05
Set up github actions

### LINKED ISSUE #1093: [4pt] Move to github actions

rosik:
With this patch, we move from Gitlab CI to Github Actions.

* Only 1.10 on ubuntu is tested for now
* Testing includes luatest (with etcd and compatibility), cypress, frontend tests, documentation, luacheck, graphql schema check 
* No pytests yet
* No mac tests yet (they were hanging, need further investigation)

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)

Close #1093


.github/actions/setup-etcd/action.yml
```diff
@@ -0,0 +1,45 @@
+name: 'Setup etcd'
+description: 'Setup tarantool environment and add it to the PATH'
```
olegrok, line 2: Seems it's typo and should be `etcd` instead of `tarantool`

.github/actions/setup-etcd/action.yml
```diff
@@ -0,0 +1,45 @@
...
+    default: false
+runs:
+  using: 'composite'
+  steps:
+    - shell: bash
+      env:
+        BASE_URL: "https://github.com/etcd-io/etcd/releases/download"
+        ETCD_VER: ${{ inputs.etcd-version }}
+        INSTALL_PREFIX: ${{ inputs.install-prefix }}
+      run: >
```
olegrok, line 22: Why do you use `<` here but `|` below? Is there any difference? If there is no difference I think it's better to uniform style

.github/actions/setup-etcd/action.yml
```diff
@@ -0,0 +1,45 @@
...
+      env:
+        INSTALL_PREFIX: ${{ inputs.install-prefix }}
+      run: |
+        set -x
+        ${INSTALL_PREFIX}/etcd --version
+        ${INSTALL_PREFIX}/etcdctl version
+
+    - shell: bash
+      run: |
+        if ( "${{ inputs.export-path }}" == "true" ); then
```
olegrok, line 41: Why do you need this "if"? Is this flow not deterministic?

.github/actions/setup-etcd/action.yml
```diff
@@ -0,0 +1,45 @@
...
+          echo "/path/to/dir" >> $GITHUB_PATH
+        fi
+
+
```
olegrok, line 45: Redundant empty lines

.github/workflows/backend-test.yml
```diff
@@ -0,0 +1,94 @@
...
+    runs-on: ${{ matrix.runs-on }}
+    env:
+      CMAKE_DUMMY_WEBUI: true
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install tarantool
+        run: |
+          curl -L https://tarantool.io/installer.sh | sudo VER=${{ matrix.tarantool }} bash
+          tarantoolctl rocks install luatest 0.5.0
```
olegrok, line 59: Seems `0.5.2` is an actual luatest version (https://github.com/tarantool/luatest/releases/tag/0.5.2)

.github/workflows/backend-test.yml
```diff
@@ -0,0 +1,94 @@
...
+          tarantoolctl rocks make
+
+      - name: Run tests
+        env:
+          ETCD_PATH: etcd-v2.3.8/etcd
+          CARTRIDGE_OLDER_PATH: cartridge-1.2.0
+        run: |
+          mkdir -p $CARTRIDGE_OLDER_PATH
+          pushd $CARTRIDGE_OLDER_PATH
+          tarantoolctl rocks install cartridge 1.2.0-1
```
olegrok, line 87: Please add a comment why do you need it

.github/workflows/frontend-test.yml
```diff
@@ -0,0 +1,54 @@
...
+    env:
+      CMAKE_DUMMY_WEBUI: false
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+
+      - name: Install tarantool
+        run: |
+          curl -L https://tarantool.io/installer.sh | sudo VER=1.10 bash
+          tarantoolctl rocks install luatest 0.5.0
```
olegrok, line 19: Ditto. 0.5.0 -> 0.5.2.

.github/workflows/frontend-test.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Run tests
+        run: |
+          ./cypress-test.sh
+          ./frontend-test.sh
+
+      - name: Cleanup cached paths
+        run: |
+          tarantoolctl rocks remove cartridge
+
```
olegrok, line 54: redundant empty line

.github/workflows/backend-test.yml
```diff
@@ -0,0 +1,94 @@
...
+          popd
+
+          .rocks/bin/luatest -v
+
+      - name: Cleanup cached paths
+        run: |
+          tarantoolctl rocks remove cartridge
```
olegrok, line 94: If the test falls, will we get to this point?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Install ldoc
+        run: >
+          tarantoolctl rocks install --server=http://rocks.moonscript.org
+          https://raw.githubusercontent.com/tarantool/ldoc/tarantool/ldoc-scm-2.rockspec
+
+      - name: Upload rockspec
+        run: |
+          TAG=${GITHUB_REF##*/}
+          ./release.sh $TAG
```
olegrok, line 45: Why do you call `rocks remove cartridge`  after tests but skip this stage here?

upload.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+URL=https://rocks.tarantool.org
+echo "Uploading $1 to $URL"
+exec curl --fail -X PUT -F "rockspec=@$1" \
+	-u "${ROCKS_USERNAME}:${ROCKS_PASSWORD}" $URL
+
```
olegrok, line 7: redundant empty line

olegrok (commented):
Thanks for your patch! Consider several comments below.

Also I would discuss several points from you description.

> Only 1.10 on ubuntu is tested for now
Testing includes luatest (with etcd and compatibility), cypress, frontend tests, documentation, luacheck, graphql schema check
No pytests yet

I think you need to create issues to fix it in future.

> No mac tests yet (they were hanging, need further investigation)

I don't see any issues as well but you have completely working MacOS CI (https://github.com/tarantool/cartridge/pull/1110) but you write "they were hanging".

.github/actions/setup-etcd/action.yml
rosik: They're different. `|` preserves newlines, while `>` replaces newlines with spaces (folds the block)

See also https://yaml-multiline.info/

.github/workflows/backend-test.yml
rosik: No, we won't. And also we won't cache anything, so it's ok.

.github/actions/setup-etcd/action.yml
rosik: You're right, it's obsolete.

.github/workflows/publish.yml
rosik: I call `rocks remove` in tests in order to cache `.rocks` directory. Here it's not necessary.

rosik:
> I think you need to create issues to fix it in future.

Here it is: #1149 

> I don't see any issues as well but you have completely working MacOS CI (#1110) but you write "they were hanging".

I can't explain it yet, I'll investigate it later

.github/actions/setup-etcd/action.yml
olegrok: Thanks for explanation. Ok.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,47 @@
...
+on:
+  push:
+    branches: [master]
+    tags:
+
+env:
+  ROCKS_USERNAME: ${{ secrets.ROCKS_USERNAME }}
+  ROCKS_PASSWORD: ${{ secrets.ROCKS_PASSWORD }}
+  CMAKE_LDOC_FIND_REQUIRED: 'YES'
+  CMAKE_SPHINX_FIND_REQUIRED: "YES"
```
olegrok, line 12: There is no difference between `"` and `'` let's support uniform style

olegrok (approved):

.github/workflows/backend-test.yml
```diff
@@ -0,0 +1,95 @@
...
+
+      - name: Lint
+        run: |
+          tarantoolctl rocks install luacheck
+          .rocks/bin/luacheck .
+
+      - name: Check schema
+        run: |
+          npm install graphql-cli@3.0.14
+          ./fetch-schema.sh
```
RunsFor, line 42: I would rather refactor steps to:

1. Install all dependencies
2. lint code
3. tarantoolctl rocks make
3. ./fetch-schema.sh

.github/workflows/backend-test.yml
```diff
@@ -0,0 +1,95 @@
...
+        uses: actions/cache@v2
+        with:
+          path: |
+            .rocks/
+            build.luarocks/
+          key: backend-build-${{ runner.os }}
+
+      - name: Build cartridge
+        run: |
+          tarantoolctl rocks make
```
RunsFor, line 78: Lets keep oneliners in one line

.github/actions/setup-etcd/action.yml
```diff
@@ -0,0 +1,34 @@
...
+runs:
+  using: 'composite'
+  steps:
+    - shell: bash
+      env:
+        BASE_URL: "https://github.com/etcd-io/etcd/releases/download"
+        ETCD_VER: ${{ inputs.etcd-version }}
+        INSTALL_PREFIX: ${{ inputs.install-prefix }}
+      run: >
+        set -x;
```
RunsFor, line 20: fyi: you can specify bash options right in the shell keyword:

```yaml
shell: bash --noprofile --norc -eox pipefail {0}
```

.github/workflows/backend-test.yml
```diff
@@ -0,0 +1,95 @@
...
+
+      - name: Run tests
+        env:
+          ETCD_PATH: etcd-v2.3.8/etcd
```
RunsFor, line 82: Why is it necessary?

.github/actions/setup-etcd/action.yml
```diff
@@ -0,0 +1,34 @@
+name: 'Setup etcd'
+description: 'Setup etcd and add it to the PATH'
```
RunsFor, line 2: Action says it adds etcd to the PATH, but I don't see it's happening.

Here is how to [add smth to system path](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#adding-a-system-path)

.github/workflows/backend-test.yml
```diff
@@ -0,0 +1,95 @@
...
+        run: >
+          tarantoolctl rocks install --server=http://rocks.moonscript.org
+          https://raw.githubusercontent.com/tarantool/ldoc/tarantool/ldoc-scm-2.rockspec
+
+      - name: Build cartridge
+        env:
+          CMAKE_LDOC_FIND_REQUIRED: 'YES'
+          CMAKE_SPHINX_FIND_REQUIRED: 'YES'
+        run: |
+          tarantoolctl rocks make
```
RunsFor, line 32: Why not extract this into `build` job and depend on it from other jobs?

- it won't duplicate
- it won't build twice for each new ref

RunsFor (changes requested):

.github/workflows/backend-test.yml
rosik: Do you want to swap linting with rocks make? What would it change?

.github/actions/setup-etcd/action.yml
rosik: Ok, but it doesn't make things simpler, so I'd rather avoid it.

.github/workflows/backend-test.yml
rosik: Because relevant luatest groups do `t.skip_if(os.getenv('ETCD_PATH') == nil, 'etcd missing')`

.github/workflows/backend-test.yml
rosik: Creating another job would require reinstalling tarantool with all deps and passing an artifact between jobs. It's not worth single-second `rocks make`.

.github/actions/setup-etcd/action.yml
rosik: Fixed

.github/workflows/backend-test.yml
RunsFor: ok. If it's not really saves time, then nevermind.

.github/workflows/backend-test.yml
RunsFor: Why not check binary in PATH instead?

.github/workflows/backend-test.yml
RunsFor: It will give you feedback faster, when it is linting error

RunsFor: and linting is quick

RunsFor (approved):

.github/actions/setup-etcd/action.yml
RunsFor: So you've decided to not setup PATH for etcd. ok then

PR #1118, opened 2020-11-11
Fix remote-control test for Tarantool 2.4+

### LINKED ISSUE #1058: [1pt] Fix remote_control_test for tarantool 2.4+

Steap2448:
Up to Tarantool 2.4 errors that come through net.box were treated as ClientError. However, since 2.4 they pass through net.box transparently preserving their error type but not the error code. That being said, remote-control behave similarly to the older versions of Tarantool and provide error codes even if there are missing. So, tests were actualized for both versions of `box.error`.

I didn't forget about

- [x] Tests
- [ ] Changelog
- [ ] Documentation

Close #1058

### DIFF
--- a/test/unit/remote_control_test.lua
+++ b/test/unit/remote_control_test.lua
@@ -530,7 +530,6 @@ function g.test_call()
         )
         t.assert_not(ok)
         t.assert_equals(type(err), 'cdata')
-        t.assert_equals(err.code, box.error.PROC_LUA)
         t.assert_equals(err.message, 'ScriptError')
 
         local ok, err = pcall(conn.call, conn, 'eval', {[1] = [[
@@ -541,7 +540,6 @@ function g.test_call()
         ]]})
         t.assert_not(ok)
         t.assert_equals(type(err), 'cdata')
-        t.assert_equals(err.code, box.error.PROC_LUA)
         t.assert_equals(err.message, 'TableError')
 
         local ok, err = pcall(conn.call, conn, 'eval',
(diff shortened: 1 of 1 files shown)


test/unit/remote_control_test.lua
```diff
@@ -530,9 +530,19 @@ function g.test_call()
         )
         t.assert_not(ok)
         t.assert_equals(type(err), 'cdata')
-        t.assert_equals(err.code, box.error.PROC_LUA)
         t.assert_equals(err.message, 'ScriptError')
 
+        -- Up to Tarantool 2.4 errors that come through net.box were
```
olegrok, line 535: Considering this comment https://github.com/tarantool/cartridge/issues/1058#issuecomment-699948095, I think it's better to remove lines that's not compatible with newest Tarantool versions. (It's a reason why I changed my opinion https://github.com/tarantool/cartridge/pull/1059#discussion_r494579391)

I think it's not useful to test that some tests don't fail on Tarantool **before** 2.4.

Steap2448: Done

olegrok:
LGTM

Maybe it could be valuable to add to comments that we still preserve netbox error contracts. Because old behaviour wasn't properly documented. (according https://github.com/tarantool/cartridge/issues/1058#issuecomment-699948095)

olegrok (approved):

rosik (approved):

Issue #587, opened 2020-02-18
Describe tarantool 1.10->2.x migration process


rosik:
Suppose i'm a user, and I have cartridge (scm-1) + tarantool (1.10). How do I upgrade tarantool to 2.x?

Related: https://github.com/tarantool/tarantool/issues/4691

- [ ] Describe the process of migration
- [ ] Choose a place for publishing this information

olegrok:
I think it could be useful - https://github.com/tarantool/cartridge/issues/253#issuecomment-538905814

PR #1125, opened 2020-11-17
gh-587 Describe tarantool 1.10->2.x migration process

Onvember:
- added info about updating schema via upgrade_schema option
- updated translation of cartridge_admin article

Close gh-587


rosik:
These two changes seems to be completely unrelated. Please, open separate PRs.

PR #1105, opened 2020-10-26
Implement roles hot-reload

rosik:
This patch allows hot-reload of cartridge roles code without restarting an instance - `cartridge.reload_roles`.

Reloading starts by stopping all roles and restoring the initial state. It's supposed that a role cleans up the global state when stopped, but even if it doesn't, cartridge kills all fibers and removes global variables and HTTP routes.

All Lua modules that were loaded during `cartridge.cfg` are unloaded, including supplementary modules required by a role. Modules, loaded before `cartridge.cfg` aren't affected.

Instance performs roles reload in a dedicated state `ReloadingRoles`. If reload fails, the instance enters the `ReloadError` state, which can later be retried. Otherwise, if reload succeeds, instance proceeds to the `ConfiguringRoles` state and initializes them as usual with `validate_config()`, `init()`, and `apply_config()` callbacks.

This is an experimental feature, it's only allowed if the application enables it explicitly: `cartridge.cfg({roles_reload_allowed = true})`.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1100

PR #1130, opened 2020-11-19
Refactor internal roles API

rosik:
This is a pre-requisite for #1105.

* Consolidate registering all roles in a single `roles.cfg()` function instead of calling `register_role()` one by one.
* Reorganize internal access to registered roles: use clear variables `by_number`, `by_package`, `by_role_name` instead of single `known_roles`. Add a layer for role meta information.

I didn't forget about

- [x] Tests
- [x] Changelog (not a public feature)

### DIFF
--- a/cartridge/roles.lua
+++ b/cartridge/roles.lua
@@ -17,108 +17,132 @@ local errors = require('errors')
 
 local vars = require('cartridge.vars').new('cartridge.roles')
 local utils = require('cartridge.utils')
-local topology = require('cartridge.topology')
 local service_registry = require('cartridge.service-registry')
 
 local RegisterRoleError = errors.new_class('RegisterRoleError')
 local ValidateConfigError = errors.new_class('ValidateConfigError')
 local ApplyConfigError = errors.new_class('ApplyConfigError')
 
-vars:new('known_roles', {
-    -- [i] = mod,
-    -- [role_name] = mod,
-})
-vars:new('roles_dependencies', {
-    -- [role_name] = {role_name_1, role_name_2}
-})
-vars:new('roles_dependants', {
-    -- [role_name] = {role_name_1, role_name_2}
-})
+vars:new('by_number', {})
+vars:new('by_package', {})
+vars:new('by_role_name', {})
 
+local function register_role(ctx, module_name)
+    checks({
+        by_number = 'table',
+        by_package = 'table',
+        by_role_name = 'table',
+    }, 'string')
 
-local function register_role(module_name)
-    checks('string')
-    local function e(...)
-        vars.known_roles = {}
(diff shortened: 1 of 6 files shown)


cartridge/roles.lua
```diff
@@ -17,108 +17,132 @@ local errors = require('errors')
...
-    -- [i] = mod,
-    -- [role_name] = mod,
-})
-vars:new('roles_dependencies', {
-    -- [role_name] = {role_name_1, role_name_2}
-})
-vars:new('roles_dependants', {
-    -- [role_name] = {role_name_1, role_name_2}
-})
+vars:new('by_number', {})
```
olegrok, line 26: To be honest I don't understand what this "by" means

cartridge/roles.lua
```diff
@@ -17,108 +17,132 @@ local errors = require('errors')
...
+vars:new('by_package', {})
+vars:new('by_role_name', {})
 
+local function register_role(ctx, module_name)
+    checks({
+        by_number = 'table',
```
olegrok, line 32: "by_number" argument is a table... what?

olegrok:
> use clear variables

it's not clear for people who read code. Please reflect the purpose of each argument.

cartridge/roles.lua
```diff
@@ -189,27 +215,22 @@ local function get_enabled_roles(roles)
...
+                end
             end
-        until true
+        end
     end
 
     return ret
 end
 
+
```
olegrok, line 232: Seems it's redundant lines

cartridge/roles.lua
```diff
@@ -17,108 +17,148 @@ local errors = require('errors')
...
-    vars.known_roles[mod.role_name] = mod
+        if not utils.table_find(role.deps, dep_role) then
+            table.insert(role.deps, dep_role)
+        end
 
-    local function deps_append(tbl, deps)
-        for _, dep in pairs(deps) do
-            if not utils.table_find(tbl, dep) then
-                table.insert(tbl, dep)
+        -- Inherit subdependencies
```
olegrok, line 98: What's about the third-level (and more deep) subdependencies?

olegrok: Seems it should be done with simple recursion function

olegrok: Or is this code really needed?

rosik: Third-level dependensies are OK because this code copies them from previous level.
It's needed and it's covered with tests: https://github.com/tarantool/cartridge/blob/2bd23ad16a028cf113a1b1ecd2317e0c6eeadfa8/test/unit/roledeps_test.lua#L133-L138

olegrok (approved):

PR #1155, opened 2020-12-03
test: failure for non-default buckets count

### LINKED ISSUE #1156: Operation error after vshard-storage role restart if non default bucket_count is used
Test https://github.com/tarantool/cartridge/pull/1155

olegrok:
It looks like vshard bug but if we don't specify "bucket_count"
explicitly we get "Non-dynamic option bucket_count cannot be reconfigured"
error. Anyway currently we should handle it on cartridge side.

Closes #1156

### DIFF
--- a/cartridge/roles/vshard-router.lua
+++ b/cartridge/roles/vshard-router.lua
@@ -94,6 +94,7 @@ local function stop()
 
     for router_name, router in pairs(vars.routers) do
         router:cfg({
+            bucket_count = router.total_bucket_count,
             sharding = {[replicaset_uuid] = {
                 replicas = {[instance_uuid] = {
                     uri = pool.format_uri(advertise_uri),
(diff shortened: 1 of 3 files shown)


olegrok:
@Gerold103 take a look one permit, please.

vshard.*.cfg fill bucket_count with default value (3000). Should we consider such behavior as vshard bug? We suppose cfg will work as box.cfg - change only specified options.

rosik (approved):

Gerold103:
Yes, it would be cool to make it work like box.cfg, but I am afraid it is too late. The current behaviour is public for quite a long time already. We can't change it.

rosik:
> Yes, it would be cool to make it work like box.cfg, but I am afraid it is too late. The current behaviour is public for quite a long time already. We can't change it.

But why?

```lua
vshard.router.cfg({bucket_count = 3001}) -- first one - ok
vshard.router.cfg({bucket_count = nil}) -- error: Non-dynamic option bucket_count cannot be reconfigured
```

I can't see a reason to throw this error.

Gerold103:
Because if I stop throwing an error here, it would be a behaviour change. Also you don't take into account other options. For example, if I omit `weight` parameter in a replicaset config, it will be set to a default without an error. If I do what you propose, the old weight will be kept. This is a breaking change, which I can't do for the existing interface.

rosik:
I've got it. Ok.

Issue #1139, opened 2020-11-25
[2pt] Revise healthcheck predicates


rosik:
Nowadays there are several different health check predicates in Cartridge.

# Eventual failover

https://github.com/tarantool/cartridge/blob/e5ca1328056bf3a3fcccc3b7a108d25d23c1f445/cartridge/failover.lua#L150-L161

# Stateful failover 

https://github.com/tarantool/cartridge/blob/e5ca1328056bf3a3fcccc3b7a108d25d23c1f445/cartridge/roles/coordinator.lua#L34-L47

# RPC

https://github.com/tarantool/cartridge/blob/e5ca1328056bf3a3fcccc3b7a108d25d23c1f445/cartridge/rpc.lua#L41-L52

I see two potential problems here:
1. Neither eventual nor stateful failover isn't triggered until a 'suspect' member becomes dead, but RPC already considers it as unhealthy. As a result, `get_active_leaders` may return a suspect leader and the RPC call would return an error "No remotes with role %q available" preliminary.
2. Stateful failover isn't triggered for OperationError, but eventual failover is. It seems to be slightly inconsistent.

rosik:
After a brief discussion with @olegrok and @mtrempoltsev we agreed that RPC should consider 'suspects' as healthy.

rosik:
And I think that OperationError shouldn't be considered unhealthy even if it sounds silly. RPC DoS and additional config applications are even worse taking into account that OperationError is the result of `apply_config` (or `init`) failure.

sharonovd:
Dmitry Sharonov, [15.06.21 14:00]
почему фейловер не сработал, я не понимаю

Dmitry Sharonov, [15.06.21 14:00]
да, мастер гикнулся при накате конфига

Yaroslav Dynnikov, [15.06.21 14:10]
https://github.com/tarantool/cartridge/issues/1139

Да, стейтфул не триггерится на OperationError.

PR #1158, opened 2020-12-04
Change RPC is_healthy predicate

rosik:
Neither eventual nor stateful failover isn't triggered until a `suspect`
member becomes dead, but RPC already considers it as unhealthy. As a
result, `get_active_leaders` may return a suspect leader and the RPC call
would return an error "No remotes with role %q available" preliminary.

Starting with this patch, prc will consider suspects as healthy.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Part of #1139

### DIFF
--- a/cartridge/rpc.lua
+++ b/cartridge/rpc.lua
@@ -42,7 +42,7 @@ local function member_is_healthy(uri, instance_uuid)
     local member = membership.get_member(uri)
     return (
         (member ~= nil)
-        and (member.status == 'alive')
+        and (member.status == 'alive' or member.status == 'suspect')
         and (member.payload.uuid == instance_uuid)
         and (
             member.payload.state == 'ConfiguringRoles' or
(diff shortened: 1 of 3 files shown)


cartridge/rpc.lua
```diff
@@ -42,7 +42,7 @@ local function member_is_healthy(uri, instance_uuid)
     local member = membership.get_member(uri)
     return (
         (member ~= nil)
-        and (member.status == 'alive')
+        and (member.status == 'alive' or member.status == 'suspect')
```
olegrok, line 45: I suggest to move such logic to some helper. To support consistency between:

- cartridge/cartridge/failover.lua
- cartridge/cartridge/roles/coordinator.lua
- cartridge/cartridge/rpc.lua

rosik: While #1139 remains unresolved (the second part) I'd rather leave it as is. Moreover, all of them have slightly different API.

olegrok (approved):

PR #1168, opened 2020-12-10
Cache frontend build intelligently

rosik:
Nowadays, the WebUI bundle is built by CMake `custom_command` (`npm ci`, `npm run build`). CMake makes decisions about rebuild basing on file timestamps. Sometimes it causes unnecessary rebuilds, e.g. when switching git branches or in CI after restoring cache.

In this patch, the WebUI build process is refactored. Instead of running `npm` directly from `custom_command`, it's wrapped into the `BuildFrontend.cmake` script. This script is executed every time, but it doesn't always run the full build. Beforehand, it calculates the checksum of target dependencies and compares it to the one persisted near build results.

The build consists of two stages:

#### 1. `npm ci`

Produces
- webui/node_modules/*

Persists checksum in
- webui/node_modules/package-lock.md5

Depends on
- webui/package-lock.json

#### 2. `npm run build`

Produces
- webui/build/bundle.lua

Persists checksum in
- webui/build/bundle.md5

Depends on
- webui/src/*
- webui/config/*.prod.js
- webui/flow-typed/*
- webui/public/*
- webui/.env
- webui/.env.production
- webui/.eslintrc
- webui/.flowconfig
- webui/codegen.yml
- webui/scripts/build.js
- webui/package-lock.json

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)
- [x] Documentation (nothing changed for a user)

### DIFF
--- a/cartridge-scm-1.rockspec
+++ b/cartridge-scm-1.rockspec
@@ -25,7 +25,6 @@ external_dependencies = {
 build = {
     type = 'cmake',
     variables = {
-        version = 'scm-1',
         TARANTOOL_DIR = '$(TARANTOOL_DIR)',
         TARANTOOL_INSTALL_LIBDIR = '$(LIBDIR)',
         TARANTOOL_INSTALL_LUADIR = '$(LUADIR)',
--- a/.github/workflows/backend-test.yml
+++ b/.github/workflows/backend-test.yml
@@ -70,7 +70,8 @@ jobs:
 
       # Run tests
       - run: .rocks/bin/luacheck .
-      - run: |
+      - name: Run tarantoolctl rocks make
+        run: |
           . ./venv/bin/activate
           tarantoolctl rocks make
         env:
--- a/.github/workflows/frontend-test.yml
+++ b/.github/workflows/frontend-test.yml
@@ -9,7 +9,11 @@ env:
 
 jobs:
   webui-test:
-    runs-on: ubuntu-18.04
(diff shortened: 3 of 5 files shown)


cartridge-scm-1.rockspec
```diff
@@ -25,7 +25,6 @@ external_dependencies = {
 build = {
     type = 'cmake',
     variables = {
-        version = 'scm-1',
```
olegrok, line 28: Why do you remove it? Seems it isn't related to your patch.

rosik: Removing this line eliminates a warning:

```text
CMake Warning:                                                                                                                                                                                             
  Manually-specified variables were not used by the project:                                                                                                                                               
                                                                                                                                                                                                           
    version    
```

olegrok:
> Avoid unnecessary frontend rebuilds basing on files checksums

This message doesn't describe anything to user who read it without reading a code. I suggest to add some notes about how it was done. You add new cmake script that manage md5 checksum of package-lock.json and another JS stuff with *.js, *.svg extensions and so on. Briefly it should be described in commit message.

cartridge-scm-1.rockspec
olegrok: Good

HustonMmmavr (approved):

olegrok (approved):

PR #1177, opened 2020-12-12
use libgraphqlparser for graphql ast build

olegrok:
Before this patch:
```lua
tarantool> require('cartridge.graphql.parse').parse('{ Test')
---
- error: Syntax error near line 1
...
```

It was completely unclear where an error occurred.

After this patch:
```lua
tarantool> require('luagraphqlparser').parse('{ Test')
---
- null
- '1.7: syntax error, unexpected EOF'
...
```

Error occured at the first line, seventh character and it's
"unexpected EOF".

But this patch breaks backward compatibility. Before this patch
empty bracets was allowd. After this patch behaviour is following:
```lua
tarantool> require('luagraphqlparser').parse('{ }')
---
- null
- '1.3: syntax error, unexpected }'
...
```

This error is returned from libgraphqlparser[1]. And seems changes
could be considered as bugfix.

Also lulpeg based graphql doesn't support inline "null" value
that's not so in libgraphqlparser.

  [1]  https://github.com/graphql/libgraphqlparser


rosik (commented):
Nice idea, I think it's worth to be developed further.

I hope you've added the submodule just for the demo purposes? Because having two rocks in the same repo isn't appropriate.

Forbidding empty brackets seems to be compliant with the GraphQL spec. I didn't find an exact paragraph (It would be nice if you could cite it), but I've found a nice online playground:

https://lucasconstantino.github.io/graphiql-online/

```json
{
  "errors": [
    {
      "message": "Syntax Error: Expected Name, found }",
      "locations": [
        {
          "line": 33,
          "column": 15
        }
      ],
      "extensions": {
        "code": "GRAPHQL_PARSE_FAILED"
      }
    }
  ]
}
```

In general, I like the opportunities of embedding libgraphqlparser. Using "locations" fields could be useful too.

olegrok:
> I hope you've added the submodule just for the demo purposes?

Yes, mainly we should choose a way how to ship this module.
Initially I thought to create a rock with libgraphqlparser submodule. But there are several reasons to implement some custom way:
  - libgraphqlparser build requires a lot of dependencies - flex, bison, python and so on. It's easier to store generated sources directly in luagraphqlparser repository (https://github.com/olegrok/luagraphqlparser). This way we require only GCC. 
 - Also default CMakeLists of libgraphqlparser doesn't allow to build it statically.

Anyway I'm think it's ready to use. It passes cartridge graphql test suite and compatibility tests with [graphql-lua](https://github.com/bjornbytes/graphql-lua) - base of cartridge graphql.

> Forbidding empty brackets seems to be compliant with the GraphQL spec. I didn't find an exact paragraph (It would be nice if you could cite it).

Yes, graphiql editor also shows error for empty brackets. But I haven't found any mentions in spec.

If you agree with moving https://github.com/olegrok/luagraphqlparser into tarantool namspace and add this rock to cartridge dependencies let's do that.

olegrok:
Will be fixed in scope of https://github.com/tarantool/cartridge/pull/1261

PR #1183, opened 2020-12-15
Return "Config didn't change" error instead of OK

### LINKED ISSUE #1046: [4pt] Cartridge skips two-phase commit if didn't change and says "OK" when it's not

<details>

My proposal:

</details>

Steap2448:
`patch_clusterwide()` now returns error "Config didn't change" error instead of OK.

I didn't forget about

- [x] Tests
- [x] Changelog
- [ ] Documentation

Close #1046


rosik:
The current design breaks backward compatibility and can't be merged.

PR #1193, opened 2020-12-18
Fix upstream.idle issue tolerance

### LINKED ISSUE #1173: [2pt] Fix upstream.idle issue tolerance
When there's no workload on a cluster, we usually see an issue: "Replication from ... to ...: long idle (1 > 1)".

The issue is shown when `upstream.idle > box.cfg.replication_timeout`.

`upstream.idle` is the time (in seconds) since the instance received the last event from a master.

<s>https://www.tarantool.io/en/doc/2.3/book/replication/repl_monitoring/

A replica sends heartbeat messages to the master every second, and the master is programmed to reconnect automatically if it does not see heartbeat messages within replication_timeout seconds.</s>

https://www.tarantool.io/en/doc/2.6/reference/configuration/#cfg-replication-replication-timeout

Since version 1.8.2. If the master has no updates to send to the replicas, it sends heartbeat messages every replication_timeout seconds, and each replica sends an ACK packet back.

Both master and replicas are programmed to drop the connection if they get no response in four replication_timeout periods. If the connection is dropped, a replica tries to reconnect to the master.

rosik:
When there's no workload on a cluster, the cartridge usually shows an issue:
"Replication from ... to ...: long idle (1 > 1)".  If the master has no
updates to send to the replicas, it sends heartbeat messages every
`replication_timeout` seconds and the issue is shown when
`upstream.idle > box.cfg.replication_timeout`.

This patch adds the 2x threshold to avoid unnecessary warnings.

I didn't forget about

- [x] Tests (not covered, hard to simulate)
- [x] Changelog
- [x] Documentation

Close #1173

### DIFF
--- a/cartridge/issues.lua
+++ b/cartridge/issues.lua
@@ -10,7 +10,7 @@
 -- * "Replication from ... to ...: high lag" -
 --   when `upstream.lag > box.cfg.replication_sync_lag`;
 -- * "Replication from ... to ...: long idle" -
---   when `upstream.idle > box.cfg.replication_timeout`;
+--   when `upstream.idle > 2 * box.cfg.replication_timeout`;
 --
 -- Failover:
 --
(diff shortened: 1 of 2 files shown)


cartridge/issues.lua
```diff
@@ -163,17 +163,18 @@ local function list_on_instance(opts)
...
-            --
-            -- Therefore, in a healthy replication setup, idle
-            -- should never exceed replication_timeout: if it does,
-            -- either the replication is lagging seriously behind,
-            -- because the master is running ahead of the replica,
-            -- or the network link between the instances is down.
+        elseif upstream.idle > 2 * box.cfg.replication_timeout then
+            -- https://www.tarantool.io/en/doc/latest/reference/configuration/#cfg-replication-replication-timeout
+
+            -- Since version 1.8.2. If the master has no updates to send to the
```
olegrok, line 169: Where did you find Tarantool 1.8?

rosik: I only cite the doc (in case the doc link rots).

rosik: Anyway, I guess it's not that important

rosik: Nevermind, I've read it again. I didn't get it right. Fixed

olegrok (approved):

PR #1194, opened 2020-12-18
Enhance UX of zone_distances configuration

### LINKED ISSUE #1096: [2pt] Enhance UX of zone_distances configuration
Related to #1086 

After a user configures zones he should also specify distances matrix. It's format isn't obvious so we should do our best to help.

Proposed solution

Let's generate `zone_distances.yml` section automatically. When a user specifies new zone, we could render the section appropriately and fill it with NULLs.

Suppose we already have `zone_distances.yml`:

```yaml
z1: {z1: 0, z2: 10}
z2: {z1: 10, z2: 0}
```

and then user makes a request:

```graphql
mutation { cluster{
  edit_topology(servers: [{uuid: "..." zone: "z3"}]){}
} }
```

It not only modifies `server.zone` property, but also updates zone_distances as follows:

```yaml
z1: {z1: 0, z2: 10, z3: null}
z2: {z1: 10, z2: 0, z3: null}
z3: {z1: null, z2: null, z3: 0}
```

rosik:
After a user configures zones he should also specify the distances matrix, but its format isn't obvious. This patch automatically creates a template for `zone_distances.yml` when a user adds the first zone. It also contains documentation references. It's not created if the section already exists.

image

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)
- [x] Documentation

Close #1096

### DIFF
--- a/cartridge/vshard-utils.lua
+++ b/cartridge/vshard-utils.lua
@@ -6,6 +6,7 @@ local errors = require('errors')
 
 local vars = require('cartridge.vars').new('cartridge.vshard-utils')
 local pool = require('cartridge.pool')
+local utils = require('cartridge.utils')
 local roles = require('cartridge.roles')
 local topology = require('cartridge.topology')
 local failover = require('cartridge.failover')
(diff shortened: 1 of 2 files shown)


cartridge/vshard-utils.lua
```diff
@@ -554,6 +555,67 @@ local function edit_vshard_options(group_name, vshard_options)
...
+
+    table.insert(template, '')
+
+    config_new:set_plaintext(
+        'zone_distances.yml',
+        table.concat(template, '\n')
+    )
+end
+
+twophase.on_patch(patch_zone_distances)
```
olegrok, line 617: It looks not hot-reload tolerant. But up to you.

olegrok (approved):

cartridge/vshard-utils.lua
rosik: I know

1. Nowadays vshard-utils isn't reloaded at all.
2. In the future we should drop all triggers (not in cartridge only, but in box too) similar to canceling fibers and removing HTTP routes.

PR #1199, opened 2020-12-22
Fix GraphQL values parsing

### LINKED ISSUE #1198: Impossible to use parseValue function for custom GraphQL scalar types
The GraphQL standard has 3 functions for handling scalar variables: parseValue, parseLiteral and serialize.
- `serialize` - the function which preparing value to send to the client
- `parseLiteral` - parsing values, which you’ve received from the client, from the query (inline values) in AST format
- `parseValue` - parsing values, which you’ve received from the client, from variables which are in JSON format (basically the same as you send them)

Now the `serialize` and `parseLiteral` functions are called correctly, but `parseValue` is never called, although it is checked during configuration. Let's call `parseValue` function during parsing to be able to use it.

Also now it's impossible to return nil values from `parseLiteral`, but all types are nullable by default.
Let's use `nil` value as error value and `box.NULL` as empty value

hackallcode:
Close #1198

### Motivation:
We want to create our own scalar type with JSON parsing. Due to the fact that the `parseValue` function was not called, it was impossible to make the correct use of variables with JSON type. So, this test fell:
https://github.com/tarantool/cartridge/blob/a39b622cd1ae9a1318e3c5d2698275241d2d2e45/test/integration/graphql_test.lua#L672-L683

### Changes:
- Now `parseValue` function is called at variables parsing step
- Now you can use `box.NULL` as return value in parsing functions `parseValue` and `parseLiteral`
- Now, if you return `nil` value from parsing functions `parseValue` and `parseLiteral`, then it is considered as a parsing error
- Now `serialize` function is called for `nil` scalar values too
- Fixed parsing of primitive scalar variables (and removed function `parseValue` from them)
- Now `parseLiteral` function as required

## Links
https://medium.com/@alizhdanov/lets-understand-graphql-scalars-3b2b016feb4a
https://atheros.ai/blog/how-to-design-graphql-custom-scalars

### DIFF
--- a/cartridge/graphql/execute.lua
+++ b/cartridge/graphql/execute.lua
@@ -189,6 +189,10 @@ local function completeValue(fieldType, result, subSelections, context, opts)
     return completedResult
   end
 
+  if fieldTypeName == 'Scalar' or fieldTypeName == 'Enum' then
+    return fieldType.serialize(result)
+  end
+
   if result == nil then
     return nil
   end
@@ -210,10 +214,6 @@ local function completeValue(fieldType, result, subSelections, context, opts)
     end
 
     return values
-  end
-
-  if fieldTypeName == 'Scalar' or fieldTypeName == 'Enum' then
-    return fieldType.serialize(result)
   end
 
   if fieldTypeName == 'Object' then
@@ -248,26 +248,24 @@ local function getFieldEntry(objectType, object, fields, context)
     argumentMap[argument.name.value] = argument
   end
 
-  local defaultValues = {}
   if context.operation.variableDefinitions ~= nil then
+    context.variables = context.variables or {}
     for _, value in ipairs(context.operation.variableDefinitions) do
-      if value.defaultValue ~= nil then
-        local variableType = query_util.typeFromAST(value.type, context.schema)
-        defaultValues[value.variable.name.value] = util.coerceValue(value.defaultValue, variableType)
+      local variable_name = value.variable.name.value
+      if type(context.variables[variable_name]) == 'nil' then
+        local default_value = box.NULL
+        if value.defaultValue ~= nil then
+          local variableType = query_util.typeFromAST(value.type, context.schema)
+          default_value = util.coerceValue(value.defaultValue, variableType)
+        end
+        context.variables[variable_name] = default_value
       end
     end
   end
 
   local arguments = util.map(fieldType.arguments or {}, function(argument, name)
     local supplied = argumentMap[name] and argumentMap[name].value
-
-    supplied = util.coerceValue(supplied, argument, context.variables,
-      {strict_non_null = true})
-    if type(supplied) ~= 'nil' then
-      return supplied
-    end
-
-    return defaultValues[name]
+    return util.coerceValue(supplied, argument, context.variables, {strict_non_null = true})
   end)
 
   --[[
--- a/cartridge/graphql/execute.lua
+++ b/cartridge/graphql/execute.lua
@@ -253,12 +253,10 @@ local function getFieldEntry(objectType, object, fields, context)
     for _, value in ipairs(context.operation.variableDefinitions) do
       local variable_name = value.variable.name.value
       if type(context.variables[variable_name]) == 'nil' then
-        local default_value = box.NULL
         if value.defaultValue ~= nil then
           local variableType = query_util.typeFromAST(value.type, context.schema)
-          default_value = util.coerceValue(value.defaultValue, variableType)
+          context.variables[variable_name] = util.coerceValue(value.defaultValue, variableType)
         end
-        context.variables[variable_name] = default_value
       end
     end
   end
--- a/cartridge/graphql/execute.lua
+++ b/cartridge/graphql/execute.lua
@@ -248,22 +248,22 @@ local function getFieldEntry(objectType, object, fields, context)
     argumentMap[argument.name.value] = argument
   end
 
+  local defaultValues = {}
   if context.operation.variableDefinitions ~= nil then
-    context.variables = context.variables or {}
     for _, value in ipairs(context.operation.variableDefinitions) do
-      local variable_name = value.variable.name.value
-      if type(context.variables[variable_name]) == 'nil' then
-        if value.defaultValue ~= nil then
-          local variableType = query_util.typeFromAST(value.type, context.schema)
-          context.variables[variable_name] = util.coerceValue(value.defaultValue, variableType)
-        end
+      if value.defaultValue ~= nil then
+        local variableType = query_util.typeFromAST(value.type, context.schema)
+        defaultValues[value.variable.name.value] = util.coerceValue(value.defaultValue, variableType)
       end
     end
   end
 
   local arguments = util.map(fieldType.arguments or {}, function(argument, name)
     local supplied = argumentMap[name] and argumentMap[name].value
-    return util.coerceValue(supplied, argument, context.variables, {strict_non_null = true})
+    return util.coerceValue(supplied, argument, context.variables, {
+      strict_non_null = true,
+      defaultValues = defaultValues,
+    })
   end)
 
   --[[
--- a/test/integration/graphql_test.lua
+++ b/test/integration/graphql_test.lua
@@ -619,9 +619,7 @@ g.test_custom_type_scalar_variables = function()
             parseLiteral = function(node)
                 return decodeJson(node.value)
             end,
-            isValueOfTheType = function(_)
-                return true
-            end,
+            isValueOfTheType = isString,
         })
 
         graphql.add_callback({
@@ -1068,8 +1066,46 @@ g.test_default_values = function()
             return args.arg.field
         end
 
+        package.loaded['test']['test_json_type'] = function(_, args)
+          if args.field == nil then
+            return nil
+          end
+          assert(type(args.field) == 'table', "Field is not a table! ")
+          assert(args.field.test ~= nil, "No field 'test' in object!")
+          return args.field
+        end
+
+        local json = require('json')
         local graphql = require('cartridge.graphql')
         local types = require('cartridge.graphql.types')
+
+        local function decodeJson(value)
+            require('log').error('$ %s', json.encode(value))
+            if value ~= nil then
+                return json.decode(value)
+            end
+            return value
+        end
+
+        local json_type = types.scalar({
+            name = 'Json',
+            description = 'Custom type with JSON decoding',
+            serialize = json.encode,
+            parseValue = decodeJson,
+            parseLiteral = function(node)
+                return decodeJson(node.value)
+            end,
+            isValueOfTheType = function(value) return type(value) == 'string' end,
+        })
+
+        graphql.add_callback({
+            name = 'test_json_type',
+            args = {
+                field = json_type,
+            },
+            kind = json_type,
+            callback = 'test.test_json_type',
+        })
 
         graphql.add_callback({
             name = 'test_default_value',
@@ -1170,5 +1206,18 @@ g.test_default_values = function()
             ]],
         variables = {arg = box.NULL}}
         ).data.test_default_object, 'nil'
+    )
+
+    t.assert_equals(
+        server:graphql({
+            query = [[
+                query($field: Json = "{\"test\": 123}") {
+                    test_json_type(
+                        field: $field
+                    )
+                }
+            ]],
+            variables = {}}
+        ).data.test_json_type, '{"test":123}'
     )
 end
--- a/test/integration/graphql_test.lua
+++ b/test/integration/graphql_test.lua
@@ -37,7 +37,7 @@ g.test_upload = function()
     server.net_box:eval([[
         package.loaded['test'] = package.loaded['test'] or {}
         package.loaded['test']['test'] = function(root, args)
-          return args[1].value
+            return args[1].value
         end
 
         local graphql = require('cartridge.graphql')
@@ -305,7 +305,7 @@ g.test_enum_input = function()
     server.net_box:eval([[
         package.loaded['test'] = package.loaded['test'] or {}
         package.loaded['test']['test_enum'] = function(root, args)
-          return args.arg.field
+            return args.arg.field
         end
 
         local graphql = require('cartridge.graphql')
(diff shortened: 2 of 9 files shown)


test/integration/graphql_test.lua
```diff
@@ -142,7 +142,7 @@ g.test_upload = function()
     )
 
     t.assert_error_msg_equals(
-        'Could not coerce "8589934592" to "Int"',
+        'Could not coerce value type "int" with value "8589934592" to type "Int"',
```
olegrok, line 145: I'm not sure it's grammatically correct. "value with value" - looks strange.

test/integration/graphql_test.lua
```diff
@@ -595,6 +625,15 @@ g.test_custom_type_scalar_variables = function()
             callback = 'test.test_custom_type_scalar',
         })
 
+        graphql.add_callback({
+            name = 'test_json_type',
+            args = {
+                field = json_type,
+            },
+            kind = json_type,
+            callback = 'test.test_custom_type_scalar',
```
olegrok, line 634: I think we need a bit different callback for tests. Currently it works as echo. But internally we should check that our variable in JSON representation was decoded to map-array-number-etc.

We should test passing this value via variables and inline specification (for primitive types only I think).

test/integration/graphql_test.lua
```diff
@@ -665,7 +743,8 @@ g.test_custom_type_scalar_variables = function()
         ).data.test_custom_type_scalar_list, 'echo'
     )
 
-    t.assert_error_msg_equals('Could not coerce "inputObject" to "CustomString"', function()
+    t.assert_error_msg_equals('Could not coerce value type "inputObject" with value "nil" ' ..
```
olegrok, line 746: Actually input value is not a string.

test/integration/graphql_test.lua
hackallcode: Maybe `Could not coerce type "int" with value "8589934592" to type "Int"` will be better

cartridge/graphql/execute.lua
```diff
@@ -248,26 +248,24 @@ local function getFieldEntry(objectType, object, fields, context)
...
-  local defaultValues = {}
   if context.operation.variableDefinitions ~= nil then
+    context.variables = context.variables or {}
     for _, value in ipairs(context.operation.variableDefinitions) do
-      if value.defaultValue ~= nil then
-        local variableType = query_util.typeFromAST(value.type, context.schema)
-        defaultValues[value.variable.name.value] = util.coerceValue(value.defaultValue, variableType)
+      local variable_name = value.variable.name.value
+      if type(context.variables[variable_name]) == 'nil' then
+        local default_value = box.NULL
```
olegrok, line 256: looks a bit strange. Are there any cases when we specify incorrect value but get null as default value?

Seems we should raise an error if variable default is not compatible with its value (for some reasons). But not silently change it to null.

test/integration/graphql_test.lua
olegrok: `Could not coerce value "8589934592"  (type: "int") to type "Int"`. However it looks really strange - "type int to Int". "8589934592" actually is not an int (but it seems to be parser issue).

test/integration/graphql_test.lua
hackallcode: This test will be failed if serialize function is not called:
https://github.com/tarantool/cartridge/blob/49349e48637e7161d48cfaf8b6bbcafbdaf92eef/test/integration/graphql_test.lua#L679-L690

This test will be failed if parseValue function is not called (if serialize was called):
https://github.com/tarantool/cartridge/blob/49349e48637e7161d48cfaf8b6bbcafbdaf92eef/test/integration/graphql_test.lua#L666-L677

This test will be failed if parseLiteral function is not called (if serialize was called):
https://github.com/tarantool/cartridge/blob/49349e48637e7161d48cfaf8b6bbcafbdaf92eef/test/integration/graphql_test.lua#L692-L703

So, I think it's not necessary to check value in callback

test/integration/graphql_test.lua
hackallcode: Yes. This test with similar error message was before this patch

cartridge/graphql/execute.lua
hackallcode: We raise an error if variable default is not compatible with its value (for some reasons). Now we set default value only if variable value is not present

test/integration/graphql_test.lua
olegrok: But it didn't show a "nil" value. Therefore it couldn't confuse user.

test/integration/graphql_test.lua
olegrok: But you pass a string and get a string. You don't check that happens inside.

`[input] -> [black box] -> [output]`

input is a string
output is a string

Change json.decode/encode in [black box] to "tostring" and you will get the same result, isn't it?

cartridge/graphql/execute.lua
olegrok: > Now we set default value only if variable value is not present

But it's wrong. It should cause an error. GraphQL has "default variable" feature and we can't simply change absence variable with null. Because in graphql it looks like:
```graphql
query($value: String = null) {  # currently our paser doesn't support null as literal but it's different kind of a problem.
  ...
}
```

test/integration/graphql_test.lua
hackallcode: Oh, yes, of course, I'll fix it

test/integration/graphql_test.lua
hackallcode: Yes, if you will use tostring function, it will be passed. But if one of parsing function not used it will be failed. But I can add check to make this test readable

olegrok: Yes, it's not enough to test an interface in this case, you need to test internals.

cartridge/graphql/execute.lua
hackallcode: Now `nil` value - is an error value. If we will use `nil` as default, then if the value of the variable is not specified, then there will be an error

rosik:
Test fails: 

```text
graphql.lua:313 E> Graphql execution failed: ./cartridge/webui/api-topology.lua:243: 'void' has no '__pairs' metamethod
```

cartridge/graphql/execute.lua
olegrok: `Now nil value - is an error value`

Was it before?
I suggest to preserve old behaviour where it's possible.
So, in case of error you can throw an error.

hackallcode: It was the same in parseLiteral

olegrok: Hm... Well, maybe we should simply remove "parseValue" function from places where it was before - to eliminate usage of then and preserve old behaviour.

Your patch should change visible behaviour for persons who implements some custom types.
But it seems not so because of - https://github.com/tarantool/cartridge/pull/1199#issuecomment-749730247 it means you start to pass box.NULL instead of nil value.

hackallcode:
> Test fails:
> 
> ```
> graphql.lua:313 E> Graphql execution failed: ./cartridge/webui/api-topology.lua:243: 'void' has no '__pairs' metamethod
> ```

I fixed `edit_topology` GraphQL function

cartridge/graphql/execute.lua
hackallcode: Okey, I make `nil` as default value of variables

test/integration/graphql_test.lua
```diff
@@ -564,25 +564,63 @@ g.test_custom_type_scalar_variables = function()
...
             parseLiteral = function(node)
-              if node.kind == 'string' then
-                return node.value
-              end
+                return decodeJson(node.value)
             end,
-            isValueOfTheType = function(value)
-              return type(value) == 'string'
+            isValueOfTheType = function(_)
+                return true
```
olegrok, line 623: Seems it should be false for non-string values

cartridge/graphql/execute.lua
```diff
@@ -248,26 +248,22 @@ local function getFieldEntry(objectType, object, fields, context)
...
-  local defaultValues = {}
   if context.operation.variableDefinitions ~= nil then
+    context.variables = context.variables or {}
     for _, value in ipairs(context.operation.variableDefinitions) do
-      if value.defaultValue ~= nil then
-        local variableType = query_util.typeFromAST(value.type, context.schema)
-        defaultValues[value.variable.name.value] = util.coerceValue(value.defaultValue, variableType)
+      local variable_name = value.variable.name.value
+      if type(context.variables[variable_name]) == 'nil' then
+        if value.defaultValue ~= nil then
```
olegrok, line 256: Maybe `if type(context.variables[variable_name]) == 'nil' and value.defaultValue ~= nil`?

test/integration/graphql_test.lua
hackallcode: Yes, but in this case it does not matter

cartridge/graphql/execute.lua
hackallcode: Yes, it's better

olegrok (approved):
Thanks for your patch. Currently it seems to be OK.

Make sure that PR description is relevant after all your changes. Also it should be great to add some sentences with motivation of this changes.

After all please ask @rosik for the second review.

test/unit/graphql_test.lua
```diff
@@ -917,33 +917,40 @@ function g.test_rules_directivesAreDefined()
...
-            end
-        end,
-        isValueOfTheType = function(value)
-            return type(value) == 'string'
+            return coerceString(node.value)
         end,
+        isValueOfTheType = isString,
     })
     t.assert_equals(CustomString.__type, 'Scalar')
 end
```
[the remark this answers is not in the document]
olegrok: > standard == spec? Seems it doesn't. Could you link some tutorial or something to help understand the internals?

https://medium.com/@alizhdanov/lets-understand-graphql-scalars-3b2b016feb4a

https://atheros.ai/blog/how-to-design-graphql-custom-scalars

Yep, it's correct to say about implementation details. It's not described in spec. But there is some examples where it's implemented. I add two links. Seems yes, we should add it into comments/commit message.

cartridge/graphql/execute.lua
```diff
@@ -189,6 +189,10 @@ local function completeValue(fieldType, result, subSelections, context, opts)
     return completedResult
   end
 
+  if fieldTypeName == 'Scalar' or fieldTypeName == 'Enum' then
+    return fieldType.serialize(result)
+  end
+
```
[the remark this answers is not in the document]
hackallcode: For example, I want to make custom type, that parse and serialize JSON objects. When I use it, I'm parsing response by `json.decode` (if it's Python code) function. If there is no `nil` serialization, then the response will contain a `nil`, but not a json string `'null'`. So, as a result, there will be a parsing error. To prevent it, I will have to use some conditions to handle `nil` separately, it's not uncomfortable. It is also logical that if we want to receive JSON at the output, then it should always be JSON (except for the case with errors).

test/integration/graphql_test.lua
```diff
@@ -564,26 +564,62 @@ g.test_custom_type_scalar_variables = function()
...
-            isValueOfTheType = function(value)
-              return type(value) == 'string'
+            isValueOfTheType = isString,
+        })
+
+        local function decodeJson(value)
+            if value ~= nil then
+                return json.decode(value)
+            end
+            return value
```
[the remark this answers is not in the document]
hackallcode: No, because `box.NULL` will be changed to `nil`

test/integration/graphql_test.lua
```diff
@@ -564,26 +564,62 @@ g.test_custom_type_scalar_variables = function()
         package.loaded['test']['test_custom_type_scalar_list'] = function(_, args)
           return args.fields[1]
         end
+        package.loaded['test']['test_json_type'] = function(_, args)
+          if args.field == nil then
+            return nil
+          end
+          assert(type(args.field) == 'table', "Field is not a table! ")
+          assert(args.field.test ~= nil, "No field 'test' in object!")
+          return args.field
```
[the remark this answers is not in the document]
hackallcode: Ok, I just wanted to use the same style as everywhere else

test/unit/graphql_test.lua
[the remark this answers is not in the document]
hackallcode: @olegrok I added links to description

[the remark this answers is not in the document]
hackallcode: > Is it tested too? Could you point it out, I didn't find it.

Yes, of course.

This is test for `box.NULL` value (`box.NULL` will be returned by `parseValue` function):
https://github.com/tarantool/cartridge/blob/a39b622cd1ae9a1318e3c5d2698275241d2d2e45/test/integration/graphql_test.lua#L685-L696

This is test for `box.NULL` value (`box.NULL` will be returned by `parseLiteral` function):
https://github.com/tarantool/cartridge/blob/a39b622cd1ae9a1318e3c5d2698275241d2d2e45/test/integration/graphql_test.lua#L698-L709

This is test for `nil` value (it will be return from `parseLiteral`):
https://github.com/tarantool/cartridge/blob/a39b622cd1ae9a1318e3c5d2698275241d2d2e45/test/integration/graphql_test.lua#L144-L151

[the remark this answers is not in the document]
hackallcode: > AFAICS the new JSON type is just a string. Seems it implies double serialization and some overhead. Is it a common way? I've seen another approach here. Could you say a few words about their pros and cons?

1. It's not a simple string, it's check availability to parse this string as JSON.
2. In your source impossible to use numbers or simple strings
3. We want to create Any type, but it's incompatible with GraphQL spec (but it's possible to implement). So, we decided that best way to solve this problem - create JSON type, that like as a string, but also have automatic JSON parsing

test/integration/graphql_test.lua
```diff
@@ -981,9 +1066,47 @@ g.test_default_values = function()
...
+        end
+
+        local json = require('json')
         local graphql = require('cartridge.graphql')
         local types = require('cartridge.graphql.types')
 
+        local function decodeJson(value)
+            require('log').error('$ %s', json.encode(value))
```
hackallcode, line 1083: Oh, it's debug info, I'll delete it

test/unit/graphql_test.lua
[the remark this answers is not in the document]
hackallcode: > What cases does this patch fix in particular? I've seen the description of #1198, could you enrich it with some examples, please.

You can see tests in previous messages, they didn't work before patch

rosik (approved):

PR #1217, opened 2021-01-12
Add implies_storage, implies_router attributes to cluster.known_roles GraphQL

### LINKED ISSUE #1071: [8pt] ☂️ WebUI usability when vshard-storage role is hidden
TDG explicitly marks vshard-storage role as hidden in favor of simple "storage". WebUI is unable to determine when vshard could be bootstrapped.

hackallcode:
Closes #1071

TDG explicitly marks vshard-storage role as hidden in favor of simple "storage". WebUI is unable to determine when vshard could be bootstrapped.

WebUI renders "Edit replicaset" dialog using the following query:
```graphql
{
  cluster {
    known_roles {
      name dependencies
    }
  }
}
```

Let's extend with `implies_storage`, `implies_router` attributes (both boolean). It may seem that `implies_router` is a new concept, but it's not. It's already used on the frontend implicitly. It's tied on a hardcoded name of the `vshard-router` role. Using this option makes it possible to eliminate implicit relations.

### DIFF
--- a/cartridge/roles/vshard-router.lua
+++ b/cartridge/roles/vshard-router.lua
@@ -239,6 +239,8 @@ end
 
 return {
     role_name = 'vshard-router',
+    implies_router = true,
+
     validate_config = vshard_utils.validate_config,
     apply_config = apply_config,
     stop = stop,
--- a/cartridge/roles/vshard-storage.lua
+++ b/cartridge/roles/vshard-storage.lua
@@ -74,6 +74,8 @@ end
 
 return {
     role_name = 'vshard-storage',
+    implies_storage = true,
+
     apply_config = apply_config,
     init = init,
     stop = stop,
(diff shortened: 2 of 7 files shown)


cartridge/roles.lua
```diff
@@ -183,22 +183,23 @@ local function get_all_roles()
...
-        and not role.M.hidden
-        then
-            table.insert(ret, role.role_name)
-        end
+        table.insert(ret, {
+            name = role.role_name,
+            hidden = role.M.permanent or role.M.hidden or false,
+            implies_router = role.M.implies_router or false,
+            implies_storage = role.M.implies_storage or false,
+        })
```
rosik, line 202: This change is incompatible.

For sure it'll break [cartridge-cli tests](https://github.com/tarantool/cartridge-cli/blob/master/test/utils.py#L865). They don't expect to see hidden roles in the list.

Changing the return format is a bad idea too. We can't be sure nobody uses it.

rosik (changes requested):

cartridge/roles.lua
hackallcode: As a result, I have added options `implies_router` and `implies_storage` in GraphQL, but not to `get_known_roles`. It should not break tests in cartridge-cli and any user's code.

rosik (approved):

PR #1239, opened 2021-01-28
Never skip two-phase commit

### LINKED ISSUE #1046: [4pt] Cartridge skips two-phase commit if didn't change and says "OK" when it's not

### LINKED ISSUE #1056: Indicate in WebUI that config apply was skipped because it didn't change
There is an undocumented feature in two-phase commit. If the config isn't changed, it's skipped:

https://github.com/tarantool/cartridge/blob/2028b0a1fd16db8c33e00b4df075716544c10a81/cartridge/twophase.lua#L259-L265

This is no good, because in this case WebUI says "OK" even if there are dead members.

image

rosik:
Commit f0b83825 introduced skipping two-phase commit if it didn't change. Back in those days, new instances used to poll the config from already bootstrapped ones, and there were lots of restrictions connected to the membership status validation. Occasionally, an attempt to apply the same config resulted in an unexpected error "Server is unreachable with status dead" despite everything was ok.

In Cartridge v2.0 everything changed. Today the bootstrap is called over remote-control, membership status checks eliminated, and updating config doesn't produce false errors anymore.

But today, we've got situations when the config is inconsistent across instances and uploading it says "ok" but does nothing.

This patch simply eliminates the skipping. It does more harm than good today.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1046 
Close #1056

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,7 +31,9 @@ Fixed
 
 - Allow specifying server zone in ``join_server`` API.
 
-- Querying zone names for dead servers.
+- Fix corner cases when querying server zones.
+
+- Don't skip config apply even if it didn't change.
 
 -------------------------------------------------------------------------------
 [2.4.0] - 2020-12-29
--- a/cartridge/twophase.lua
+++ b/cartridge/twophase.lua
@@ -330,14 +330,6 @@ local function _clusterwide(patch)
 
     topology.probe_missing_members(topology_new.servers)
 
-    if utils.deepcmp(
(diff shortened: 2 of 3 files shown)


olegrok:
> But today, we've got situations when the config is inconsistent across instances and uploading it says "ok" but does nothing.

An alternative solution is to compare config checksums with another instances and if all of them are the same still do nothing.
Configuration apply is too dangerous.

rosik:
Rejected

PR #1245, opened 2021-02-01
Add suggestion to disable malfunctioning servers

### LINKED ISSUE #1068: [2pt] Suggest to disable offline instances
Similarly to #1036 we could fix another case - restore the quorum after a part of cluster goes offline.

The specifics of two-phase commit doesn't let you to expel a server if there is two or more offline. While one request tries to expel one of them, the others prevent successfully applying  two-phase commit. As a result, "Expel server" button becomes useless in this case.

With a disable suggestion we could restore the quorum with a single button click. And user will decide later whether to repair and re-enable it or to expel completely.

Related to #50

Steap2448:
New suggestion to disable malfunctioning servers in order to restore the quorum:

```GraphQl
query {
  cluster {
    suggestions {
      disable_servers  {
        uuid
      } 
    }
  }
}
```

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1068

### DIFF
--- a/cartridge/webui/api-suggestions.lua
+++ b/cartridge/webui/api-suggestions.lua
@@ -80,7 +80,7 @@ local function force_apply(_, _, info)
 
     local cache = info.context.request_cache
     if cache.issues == nil then
-        cache.issues = issues.list_on_cluster()
+        cache.issues, cache.issues_err = issues.list_on_cluster()
     end
 
     local reasons_map = {}
@@ -126,7 +126,7 @@ local function force_apply(_, _, info)
     return ret
 end
 
-local function disable_servers()
+local function disable_servers(_, _, info)
     local topology_cfg = confapplier.get_readonly('topology')
     if topology_cfg == nil then
         return nil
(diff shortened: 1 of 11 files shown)


cartridge/webui/api-suggestions.lua
```diff
@@ -36,6 +36,16 @@ local force_apply_suggestion = gql_types.object({
     }
 })
 
+local disable_servers_suggestion = gql_types.object({
+    name = 'DisableServersSuggestion',
```
rosik, line 40: Let's call it "DisableServerSuggestion"? Single, not plural.

cartridge/webui/api-suggestions.lua
```diff
@@ -116,10 +126,44 @@ local function force_apply(_, _, info)
...
+    end
+
+    if cache.issues_err == nil then
+        return nil
+    end
+
+    local refined_uri_list = topology.refine_servers_uri(topology_cfg)
+    for _, uuid, _ in fun.filter(topology.not_disabled, topology_cfg.servers) do
+        local member = membership.get_member(refined_uri_list[uuid])
+
```
rosik, line 149: Redundant. And what if it's not in membership?

doc/schema.graphql
```diff
@@ -68,6 +68,13 @@ type DDLSchema {
   as_yaml: String!
 }
 
+"""
+A suggestion to disable malfunctioning servers  in order to restore the quorum
```
rosik, line 72: Excess whitespace: `servers  in`

rosik (approved):

PR #1250, opened 2021-02-03
Fix version generation upon publishing

### LINKED ISSUE #1214: [2pt] Fix version generation in CI
Release 2.4.0 published with CI did contain `VERSION = "unknown"`. I've fixed it manually, but it also should be fixed (and tested) in CI.

rosik:
Luarocks command `install` builds a rock in a temp directory and removes
any source artifacts (including .git directory). Without it, cmake is
unable to detect the version and released `all.rock` states the version
"unknown". Since there is no other source of version info this patch
replaces the `install` command with `make`.

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)

Close #1214

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -38,7 +38,20 @@ jobs:
       # Make a release
       - run: echo TAG=${GITHUB_REF##*/} >> $GITHUB_ENV
       - run: tarantoolctl rocks new_version --tag ${{ env.TAG }}
-      - run: tarantoolctl rocks install cartridge-${{ env.TAG }}-1.rockspec
+      - run: tarantoolctl rocks make cartridge-${{ env.TAG }}-1.rockspec
+      - run: |
+          mkdir -p check-version
+          # get away from repo sources because `cartridge/VERSION.lua`
+          # always returns 'scm-1' and assertion fails
+          pushd check-version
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -38,7 +38,19 @@ jobs:
       # Make a release
       - run: echo TAG=${GITHUB_REF##*/} >> $GITHUB_ENV
       - run: tarantoolctl rocks new_version --tag ${{ env.TAG }}
-      - run: tarantoolctl rocks install cartridge-${{ env.TAG }}-1.rockspec
+      - run: tarantoolctl rocks make cartridge-${{ env.TAG }}-1.rockspec
+      - run: |
+          mkdir -p check-version
+          # make sure the modules are loaded from ../.rocks, but not ./
```
olegrok, line 44: Also you could verify it with "package.search('cartridge')"

olegrok (approved):

.github/workflows/publish.yml
rosik: I guess the words "make sure" aren't accurate enough here. It's not necessary to check `package.search` because if it's `./cartridge.lua`, then it'll always return 'scm-1' from `cartridge/VERSION.lua` (which isn't rendered during build) and thus the assertion will always fail.

So I've changed the comment.

PR #1254, opened 2021-02-04
Fix flaky ddl.test_space_removal test

rosik:
The test applies a cluster-wide config with a modified schema and checks the space existence on all servers. But the replication is asynchronous and the check sometimes fails on a replica. This patch wraps it with `helpers.retrying`.

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)

### DIFF
--- a/test/integration/ddl_test.lua
+++ b/test/integration/ddl_test.lua
@@ -198,10 +198,12 @@ function g.test_space_removal()
     _set_schema(server, 'spaces: {}')
 
     for _, srv in pairs(g.cluster.servers) do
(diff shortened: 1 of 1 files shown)


olegrok (approved):
Does helpers.retrying have any default timeout?
Maybe the most proper way is to use wait_lsn. But feel free to ignore - for tests it's good enough.

rosik:
> Does helpers.retrying have any default timeout?

Yes, it does. 5 sec. It's the briefest way for testing.

PR #1260, opened 2021-02-10
Add bootstrap proxy calls

### LINKED ISSUE #1244: [4pt] Bootstrap cluster instance remotely
Consider a case. There are two instances. One is already bootstrapped and the second one isn't.

WebUI on the first one looks as usual:

image

But the second one looks confusing

image

Moreover, if a user bootstraps the second one, he'll get two independent clusters.

Unconfigured instances should proxy calls to the bootstrapped ones if there are any.

HustonMmmavr:
* Added proxy calls to configured instances for:
  - botstrap_vshard
  - edit_topology
  - get_topology
* Added new methods at cartridge.rpc

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1244


rosik:
Closing in favor of #1278

PR #1266, opened 2021-02-15
Make clusterwide config upload a separate stage

### LINKED ISSUE #1145: [2pt] Uploading big config fails on large clusters
When the cluster-wide configuration is updated, Cartridge sends it whole to every instance in a cluster. He does it in parallel and imposes a hardcoded 5-second timeout restriction. If we suppose that there are 100 instances, the size of config is 1MB and the network bandwidth is 100Mb/s, then it would take 8 seconds and obviously fail.

Steap2448:
For historical reasons, there were two issues in the `patch_clusterwide` logics:

1. The 5-second timeout for all communications was hardcoded.
2. No optimizations were undertaken there, it was was suboptimal and wasteful.

It was OK when the config was small and there were only a few instances in the cluster, but it's a problem for big installations nowadays.

This patch introduces a new `patch_clusterwide` stage - upload (consolidated in a separate module). It spreads the data across instances in a network-efficient manner.

Also in this patch:

- Change `issues_test.lua` and remove unnecessary checks from cypress `sigstop.spec.js` because the SIGStOP doesn't result in config lock and mismatch anymore.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1145

### DIFF
--- a/cartridge/upload.lua
+++ b/cartridge/upload.lua
@@ -61,6 +61,8 @@ local function upload_begin(upload_id)
 
     local upload_path = get_upload_path(upload_id)
     local ok, err = utils.mktree(fio.dirname(upload_path))
+    -- If the fiber is cancelled, the `upload_fibers` is cleaned up
+    -- during the `upload_cleanup` stage (which in fact cancelled it).
     fiber.testcancel()
     if ok == nil then
         vars.upload_fibers[upload_id] = nil
@@ -96,6 +98,8 @@ local function upload_transmit(upload_id, payload)
     local upload_path = get_upload_path(upload_id)
     local payload_path = fio.pathjoin(upload_path, 'payload')
     local ok, err = utils.file_write(payload_path, payload)
+    -- If the fiber is cancelled, the `upload_fibers` is cleaned up
+    -- during the `upload_cleanup` stage (which in fact cancelled it).
     fiber.testcancel()
     if not ok then
         vars.upload_fibers[upload_id] = nil
--- a/cartridge/upload.lua
+++ b/cartridge/upload.lua
@@ -7,7 +7,6 @@
 
 local log = require('log')
 local fio = require('fio')
-local fun = require('fun')
 local errno = require('errno')
 local fiber = require('fiber')
 local checks = require('checks')
@@ -32,18 +31,24 @@ vars:new('inbox', {})
 -- before previous stages finish.
 vars:new('upload_fibers', {})
 
-vars:new('options', {
-    upload_prefix = '/tmp',
-    netbox_call_timeout = 1,
-    transmission_timeout = 30,
-})
+-- Temporary directory used for saving files during upload.
+vars:new('upload_prefix', '/tmp')
 
-vars:new('get_upload_path', function(upload_id)
+local function set_upload_prefix(prefix)
+    vars.upload_prefix = prefix
+end
+
+local function get_upload_prefix()
+    return vars.upload_prefix
+end
+
+local function get_upload_path(upload_id)
+    checks('string')
     return fio.pathjoin(
-        vars.options.upload_prefix,
+        vars.upload_prefix,
         'cartridge-upload.' .. upload_id
     )
-end)
+end
 
 -- The upload starts by creating a shared resource - a directory named
 -- after the `upload_id`. The instances who can create it before
@@ -54,7 +59,7 @@ local function upload_begin(upload_id)
 
     vars.upload_fibers[upload_id] = fiber.self()
 
-    local upload_path = vars.get_upload_path(upload_id)
+    local upload_path = get_upload_path(upload_id)
     local ok, err = utils.mktree(fio.dirname(upload_path))
     fiber.testcancel()
     if ok == nil then
@@ -89,7 +94,7 @@ local function upload_transmit(upload_id, payload)
 
     vars.upload_fibers[upload_id] = fiber.self()
 
-    local upload_path = vars.get_upload_path(upload_id)
+    local upload_path = get_upload_path(upload_id)
     local payload_path = fio.pathjoin(upload_path, 'payload')
     local ok, err = utils.file_write(payload_path, payload)
     fiber.testcancel()
@@ -109,7 +114,7 @@ local function upload_finish(upload_id)
 
     vars.upload_fibers[upload_id] = fiber.self()
 
-    local upload_path = vars.get_upload_path(upload_id)
+    local upload_path = get_upload_path(upload_id)
     local payload_path = fio.pathjoin(upload_path, 'payload')
     local payload, err = utils.file_read(payload_path)
 
@@ -136,7 +141,7 @@ local function upload_cleanup(upload_id)
         vars.upload_fibers[upload_id] = nil
     end
 
-    local upload_path = vars.get_upload_path(upload_id)
+    local upload_path = get_upload_path(upload_id)
     local random_path = utils.randomize_path(upload_path)
 
     local ok = fio.rename(upload_path, random_path)
@@ -188,8 +193,12 @@ end
 -- @treturn[1] string `upload_id` (if at least one upload succeded)
 -- @treturn[2] nil
 -- @treturn[2] table Error description
-local function upload(data, uri_list)
-    checks('?', 'table')
+local function upload(data, opts)
+    checks('?', {
+        uri_list = 'table',
+        netbox_call_timeout = '?number',
+        transmission_timeout = '?number',
+    })
     local ok, payload = pcall(msgpack.encode, data)
     if not ok then
         return nil, UploadError:new(
@@ -205,12 +214,12 @@ local function upload(data, uri_list)
         local retmap, errmap = pool.map_call(
             '_G.__cartridge_upload_begin', {upload_id},
             {
-                uri_list = uri_list,
-                timeout = vars.options.netbox_call_timeout,
+                uri_list = opts.uri_list,
+                timeout = opts.netbox_call_timeout,
             }
         )
 
-        for _, uri in ipairs(uri_list) do
+        for _, uri in ipairs(opts.uri_list) do
             if retmap == nil or retmap[uri] == nil then
                 local err = errmap and errmap[uri]
                 if err == nil then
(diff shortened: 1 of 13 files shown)


HustonMmmavr (approved):

cartridge/upload.lua
```diff
@@ -0,0 +1,313 @@
...
+local function upload_begin(upload_id)
+    checks('string')
+
+    vars.upload_fibers[upload_id] = fiber.self()
+
+    local upload_path = vars.get_upload_path(upload_id)
+    local ok, err = utils.mktree(fio.dirname(upload_path))
+    fiber.testcancel()
+    if ok == nil then
+        return nil, err
```
rosik, line 65: cleanup vars.upload_fibers

cartridge/upload.lua
```diff
@@ -0,0 +1,313 @@
...
+
+_G.__cartridge_upload_begin = function(...) return errors.pcall('E', upload_begin, ...) end
+_G.__cartridge_upload_transmit = function(...) return errors.pcall('E', upload_transmit, ...) end
+_G.__cartridge_upload_finish = function(...) return errors.pcall('E', upload_finish, ...) end
+_G.__cartridge_upload_cleanup = function(...) return errors.pcall('E', upload_cleanup, ...) end
+
+return {
+    upload = upload,
+    inbox = vars.inbox,
+    _options = vars.options,
```
rosik, line 312: Rename, make configurable via argparse

cartridge/upload.lua
```diff
@@ -0,0 +1,313 @@
...
+--
+-- @param data
+--   any Lua object.
+-- @tparam {string,...} uri_list
+--   array of URIs.
+--
+-- @treturn[1] string `upload_id` (if at least one upload succeded)
+-- @treturn[2] nil
+-- @treturn[2] table Error description
+local function upload(data, uri_list)
```
olegrok, line 193: This module is quite atomic and can be used by other projects, let's move timeout to options

olegrok: `upload_path ` also should be moved to options. I think all constants/options could be passed via arguments.
Cartridge should provide only default values and do it out of the scope this module.

cartridge/upload.lua
```diff
@@ -0,0 +1,313 @@
...
+local utils = require('cartridge.utils')
+
+local vars = require('cartridge.vars').new('cartridge.upload')
+
+local UploadError = errors.new_class('UploadError')
+
+--- The uploaded data.
+-- `{[upload_id] = data}`
+-- @table inbox
+vars:new('inbox', {})
```
olegrok, line 26: It could be managed externally. Maybe as a Tarantool future.
`upload` could result some object that has some condition. We could wait result, interrupt uploading, etc via such object.
Also it should be user responsibility how to manage result of upload. And it solves a problem when in case of failure we store garbage on filesystem - `__gc` metamethod or `cleanup` callback should remove this garbage.

cartridge/upload.lua
rosik: > This module is quite atomic and can be used by other projects, let's move timeout to options

Good idea. Done.



> `upload_path ` also should be moved to options. I think all constants/options could be passed via arguments.
> Cartridge should provide only default values and do it out of the scope this module.

And this seems to be impossible. The sending side can't know where the accepting side wants to store the data.

cartridge/upload.lua
rosik: It's a distributed system. The uploaded data can't be deleted before the sender makes another netbox call (`prepare_2pc`). The only thing I could do is a time-managed GC, but it's not a cure - overcomplicated and there still may be problems.

cartridge/upload.lua
```diff
@@ -0,0 +1,322 @@
...
+-- others are the transmitters. The others wait patiently till the
+-- `upload_finish()`.
+local function upload_begin(upload_id)
+    checks('string')
+
+    vars.upload_fibers[upload_id] = fiber.self()
+
+    local upload_path = get_upload_path(upload_id)
+    local ok, err = utils.mktree(fio.dirname(upload_path))
+    fiber.testcancel()
```
olegrok, line 64: `testcancel` raises. You won't perform `vars.upload_fibers[upload_id] = nil` if this function raises an error.

cartridge/upload.lua
```diff
@@ -0,0 +1,322 @@
...
+-- during the `upload_finish()` step.
+local function upload_transmit(upload_id, payload)
+    checks('string', 'string')
+
+    vars.upload_fibers[upload_id] = fiber.self()
+
+    local upload_path = get_upload_path(upload_id)
+    local payload_path = fio.pathjoin(upload_path, 'payload')
+    local ok, err = utils.file_write(payload_path, payload)
+    fiber.testcancel()
```
olegrok, line 100: `testcancel` raises. You won't perform `vars.upload_fibers[upload_id] = nil` if this function raises an error.

cartridge/upload.lua
```diff
@@ -0,0 +1,322 @@
...
+    if vars.upload_fibers[upload_id] ~= nil then
+        pcall(fiber.cancel, vars.upload_fibers[upload_id])
+        vars.upload_fibers[upload_id] = nil
+    end
+
+    local upload_path = get_upload_path(upload_id)
+    local random_path = utils.randomize_path(upload_path)
+
+    local ok = fio.rename(upload_path, random_path)
+    local _errno = errno()
```
olegrok, line 148: It's redundant action to call errno here. It's not needed if ok is true.

cartridge/upload.lua
```diff
@@ -0,0 +1,322 @@
...
+                'Error removing %s: %s',
+                upload_path, errno.strerror(_errno)
+            )
+            return false
+        end
+    end
+
+    local ok = fio.rmtree(random_path)
+    local _errno = errno()
+    if not ok then
```
olegrok, line 164: It's redundant action to call errno here. It's not needed if ok is true.

cartridge/upload.lua
```diff
@@ -0,0 +1,322 @@
...
+        return nil, _upload_error
+    end
+
+    return upload_id
+end
+
+_G.__cartridge_upload_begin = function(...) return errors.pcall('E', upload_begin, ...) end
+_G.__cartridge_upload_transmit = function(...) return errors.pcall('E', upload_transmit, ...) end
+_G.__cartridge_upload_finish = function(...) return errors.pcall('E', upload_finish, ...) end
+_G.__cartridge_upload_cleanup = function(...) return errors.pcall('E', upload_cleanup, ...) end
```
olegrok, line 314: You don't actually need to pass `...` here because you exactly know a number of arguments - one upload_id.
varargs are bit slower than usual function call with determined number of arguments.

test/integration/upload_test.lua
```diff
@@ -0,0 +1,161 @@
...
+        err = 'Timeout exceeded',
+    })
+
+    -- prefix should be cleaned up even if upload_begin fails
+    t.assert_equals(fio.listdir(g.upload_path_1), {})
+    t.assert_equals(fio.listdir(g.upload_path_2), {})
+
+    g.s2.process:kill('CONT')
+    -- give the instance some time to wake up after stop
+    fiber.sleep(0.3)
```
olegrok, line 75: Could you replace it with some wait_for?

test/integration/upload_test.lua
```diff
@@ -0,0 +1,161 @@
...
+        server_command = helpers.entrypoint('srv_basic'),
+        cookie = helpers.random_cookie(),
+        replicasets = {{
+            alias = 'main',
+            uuid = helpers.uuid('a'),
+            roles = {},
+            servers = {
+                -- first upload group: s1 and s2
+                {env = {TARANTOOL_UPLOAD_PREFIX = g.upload_path_1}},
+                {env = {TARANTOOL_UPLOAD_PREFIX = g.upload_path_1}},
```
olegrok, line 25: Maybe it's reasonable to increase number of servers up to 5 or even more. In case of problems with concurrency and shared resource problems we have more chances to catch a problem in CI.

cartridge/upload.lua
rosik: It's ok. In this case the `upload_fibers` will be cleaned by the `cleanup` stage which cancelled the fiber.

olegrok: Ok, but it's not so obvious, please add a comment.

olegrok: However, in fact we can't guarantee that cleanup stage will be executed someday in case of temporary problems with network.

cartridge/upload.lua
rosik: The main danger of errno is that it may be overridden by some other consequent fio operation. I've changed this place, but in another way.

test/integration/upload_test.lua
rosik: I was going to, but the predicate turned out to be super complicated. I need to wait when another async non-atomic (yielding) netbox call finishes on the server-side. I've decided that sleep is the shortest way. If the test turns out to be flaky I'll fix it later, but today it seems to be ok.

cartridge/upload.lua
rosik: > Ok, but it's not so obvious, please add a comment.

Done

> However, in fact we can't guarantee that cleanup stage will be executed someday in case of temporary problems with network.

The necessity of persisting `upload_fibers` is to avoid a race between `begin` and `cleanup`. The necessity of cleaning up the `upload_fibers` is to avoid canceling an innocent fiber that belongs to the iproto fiber pool and has already been reused by that time.

cartridge/upload.lua
olegrok:
```lua
local _errno
local ok, err = fio.some_operation()
if not ok then
   _errno = errno()
end
```

No redefinition, to excess actions.

test/integration/upload_test.lua
olegrok: ok

cartridge/upload.lua
rosik: To sum up, if the cleanup stage wasn't called for some reason, the garbage will remain in `/tmp` and in lua context. And it's ok for now.

cartridge/upload.lua
rosik: I'm ok to sacrifice that performance for the sake of code brevity. Hope it's not too much.

olegrok (approved):

rosik (approved):

Issue #1274, opened 2021-02-25
[4pt] Fix problems with slow webui requests


rosik:
Steps to reproduce: 

0. bootstrap a 2-instance cluster and stop it.
1. start srv-2
2. send a sigstop to srv-2
3. start srv-1.
4. open webui on srv-1

It'll take about 20sec to load. Expected is 1-2 sec.

PR #1278, opened 2021-03-01
Proxy API calls from and an unconfigured instance

### LINKED ISSUE #1244: [4pt] Bootstrap cluster instance remotely

rosik:
Nowadays, every unconfigured instance only displays itself and the other unconfigured instances. Even if there's a cluster somewhere near, it isn't shown.

With this patch, a user is able to observe the cluster even from an unconfigured server. The only requirement is that it's connected to the membership.

image

Notice, that WebUI is open on the unconfigured instance.

There're still some restrictions regarding managing the cluster. One can't edit failover params, auth and users, code and scheme, upload / download config. But the current unconfigured instance can be joined to the existing cluster. And it doesn't result in another separate cluster creation anymore.

This is the alternative implementation of #1260. The proxy calls are implemented in a more compact manner (and in a separate module).

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1244

### DIFF
--- a/cartridge/issues.lua
+++ b/cartridge/issues.lua
@@ -47,18 +47,22 @@
 --
 -- @module cartridge.issues
 -- @local
+local mod_name = 'cartridge.issues'
 
 local fio = require('fio')
 local log = require('log')
 local fun = require('fun')
 local fiber = require('fiber')
 local checks = require('checks')
+local membership = require('membership')
+
 local pool = require('cartridge.pool')
 local topology = require('cartridge.topology')
-local confapplier = require('cartridge.confapplier')
 local failover = require('cartridge.failover')
-local membership = require('membership')
-local vars = require('cartridge.vars').new('cartridge.issues')
+local confapplier = require('cartridge.confapplier')
+local lua_api_proxy = require('cartridge.lua-api.proxy')
+
+local vars = require('cartridge.vars').new(mod_name)
 
 --- Thresholds for issuing warnings.
 -- All settings are local, not clusterwide. They can be changed with
@@ -337,6 +341,13 @@ local function list_on_instance(opts)
 end
 
 local function list_on_cluster()
+    local ret, err = lua_api_proxy.call(mod_name .. '.list_on_cluster')
+    if ret ~= nil then
+        return ret
+    elseif err ~= nil then
+        return nil, err
+    end
+
     local ret = {}
     local uri_list = {}
     local topology_cfg = confapplier.get_readonly('topology')
--- a/cartridge/issues.lua
+++ b/cartridge/issues.lua
@@ -341,10 +341,16 @@ local function list_on_instance(opts)
 end
 
 local function list_on_cluster()
-    local ret, err = lua_api_proxy.call(mod_name .. '.list_on_cluster')
-    if ret ~= nil then
-        return ret
-    elseif err ~= nil then
+    local state, err = confapplier.get_state()
+    if state == 'Unconfigured' and lua_api_proxy.can_call()  then
+        -- Try to proxy call
+        local ret = lua_api_proxy.call(mod_name .. '.list_on_cluster')
+        if ret ~= nil then
+            return ret
+        -- else
+            -- Don't return an error, go on
+        end
+    elseif state == 'InitError' or state == 'BootError' then
         return nil, err
     end
 
--- a/cartridge/lua-api/get-topology.lua
+++ b/cartridge/lua-api/get-topology.lua
@@ -1,6 +1,7 @@
 --- Administration functions (`get-topology` implementation).
 --
 -- @module cartridge.lua-api.get-topology
+local mod_name = 'cartridge.lua-api.get-topology'
 
 local fun = require('fun')
 local membership = require('membership')
@@ -10,6 +11,8 @@ local roles = require('cartridge.roles')
 local failover = require('cartridge.failover')
 local topology = require('cartridge.topology')
 local confapplier = require('cartridge.confapplier')
+
+local lua_api_proxy = require('cartridge.lua-api.proxy')
 
 --- Replicaset general information.
 -- @tfield string uuid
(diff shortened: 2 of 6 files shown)


cartridge/lua-api/proxy.lua
```diff
@@ -0,0 +1,66 @@
...
+
+    for _, member in membership.pairs() do
+        if member.status ~= 'alive'
+        or member.payload.uuid == nil
+        then
+            goto continue
+        end
+
+        if vars.destination == nil
+        or member.uri < vars.destination
```
HustonMmmavr, line 36: Add commentary

cartridge/issues.lua
```diff
@@ -337,6 +341,13 @@ local function list_on_instance(opts)
 end
 
 local function list_on_cluster()
+    local ret, err = lua_api_proxy.call(mod_name .. '.list_on_cluster')
+    if ret ~= nil then
+        return ret
+    elseif err ~= nil then
+        return nil, err
+    end
```
olegrok, line 349: Looks a bit tricky. If remote function returns "box.NULL" or nothing it is considered as need to perform a "local call".

cartridge/lua-api/proxy.lua
```diff
@@ -0,0 +1,67 @@
...
+
+    if confapplier.get_state() ~= 'Unconfigured' then
+        return nil
+    end
+
+    if vars.destination ~= nil then
+        local member = membership.get_member(vars.destination)
+        if member ~= nil and member.status == 'alive' then
+            goto call
+        end
```
olegrok, line 24: `else vars.destination = nil`?

cartridge/lua-api/proxy.lua
```diff
@@ -0,0 +1,67 @@
...
+    if vars.destination == nil then
+        return nil
+    end
+
+::call::
+    local conn = pool.connect(vars.destination, {wait_connected = false})
+    local ret, err = errors.netbox_eval(conn, [[
+        local ret, err = require('cartridge.funcall').call(...)
+        if ret == nil then return nil, err end
+        return require('yaml').encode(ret)
```
olegrok, line 54: is yaml needed here as protection from recursive structures?

Please add a comment.

cartridge/lua-api/proxy.lua
```diff
@@ -0,0 +1,67 @@
...
+        ::continue::
+    end
+
+    if vars.destination == nil then
+        return nil
+    end
+
+::call::
+    local conn = pool.connect(vars.destination, {wait_connected = false})
+    local ret, err = errors.netbox_eval(conn, [[
```
olegrok, line 51: It it possible that you will perform a call before connection will be established?

cartridge/lua-api/proxy.lua
```diff
@@ -0,0 +1,67 @@
...
+        local ret, err = require('cartridge.funcall').call(...)
+        if ret == nil then return nil, err end
+        return require('yaml').encode(ret)
+    ]], {function_name, ...})
+
+    if ret == nil then
+        vars.destination = nil
+        return nil, err
+    end
```
olegrok, line 60: Maybe retry? If there are other members

cartridge/lua-api/proxy.lua
rosik: Yes, the connection internal state machine handles it fine. Moreover, it's the only valid way to avoid long timeouts (like in #1274)

rosik:
We've refactored the patch and the API

cartridge/lua-api/proxy.lua
rosik: Fixed

cartridge/lua-api/proxy.lua
rosik: If something goes wrong it would be terrible in a large cluster. Better fall back to the old behavior.

cartridge/lua-api/proxy.lua
rosik: Done

cartridge/issues.lua
rosik: Fixed.

cartridge/lua-api/proxy.lua
olegrok: Agree

cartridge/issues.lua
```diff
@@ -337,6 +341,19 @@ local function list_on_instance(opts)
...
 
 local function list_on_cluster()
+    local state, err = confapplier.get_state()
+    if state == 'Unconfigured' and lua_api_proxy.can_call()  then
+        -- Try to proxy call
+        local ret = lua_api_proxy.call(mod_name .. '.list_on_cluster')
+        if ret ~= nil then
+            return ret
+        -- else
+            -- Don't return an error, go on
```
olegrok, line 351: Is it covered with tests somehow?

cartridge/lua-api/get-topology.lua
```diff
@@ -65,8 +68,15 @@ local confapplier = require('cartridge.confapplier')
...
-    if state == 'InitError' or state == 'BootError' then
+    if state == 'Unconfigured' and lua_api_proxy.can_call() then
+        -- Try to proxy call
+        local ret = lua_api_proxy.call(mod_name .. '.get_topology')
+        if ret ~= nil then
+            return ret
+        -- else
+            -- Don't return an error, go on
+        end
+    elseif state == 'InitError' or state == 'BootError' then
```
olegrok, line 79: I see similar code at least two times. Maybe it's reasonable to move it to separate function?

olegrok (approved):

PR #1279, opened 2021-03-01
Cluster option to disable webui

### LINKED ISSUE #551: Disable the admin UI by default
Currently the admin UI is enabled by default. But many users in traditional scenarios may not need it. Let's add a parameter to cartridge.cfg like `admin_gui_enabled` and set it to false by default.

Steap2448:
New `cartridge.cfg` option `webui_enabled` (default: true). Otherwise, HTTP server remains operable (and GraphQL too), but serves user-defined roles API only.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #551

### DIFF
--- a/cartridge/webui.lua
+++ b/cartridge/webui.lua
@@ -30,7 +30,6 @@ local function init(httpd)
     front.init(httpd)
     front.add('cluster', front_bundle)
 
-    graphql.init(httpd)
     graphql.add_mutation_prefix('cluster', 'Cluster management')
     graphql.add_callback_prefix('cluster', 'Cluster management')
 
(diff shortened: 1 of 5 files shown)


cartridge/webui.lua
```diff
@@ -26,9 +26,11 @@ local function get_blacklist()
...
-local function init(httpd)
-    front.init(httpd)
-    front.add('cluster', front_bundle)
+local function init(httpd, webui_enabled)
+    if webui_enabled then
+        front.init(httpd)
+        front.add('cluster', front_bundle)
+    end
 
     graphql.init(httpd)
```
rosik, line 35: Let's move this single line to `cartridge.lua`. The logic will be the following.

1. If http_enabled, we start httpd and call graphql.init(httpd)
2. If webui_enabled, we call  front.init(httpd) + all graphql.add_callback / add_mutation

rosik (approved):

PR #1285, opened 2021-03-03
Sandbox unit.cfg_errors test

rosik:
Recently we've encountered a weird test behavior: a test (`unit.upload` from #1266) succeded when run alone, but failed in conjunction with `unit.cfg_errors` test. The reason is in monkeypatching and reloading of lua modules used there.

With this patch, all test functions are moved to a simple empty server and run over netbox (remote-control).

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)

### DIFF
--- a/test/helper.lua
+++ b/test/helper.lua
@@ -1,8 +1,10 @@
 require('strict').on()
 
 local fio = require('fio')
+local checks = require('checks')
 local digest = require('digest')
 local helpers = table.copy(require('cartridge.test-helpers'))
+local utils = require('cartridge.utils')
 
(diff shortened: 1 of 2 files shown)


olegrok (approved):
Please file at least an issue to luatest to have an ability to run unit tests in separate tarantool process. Currently it looks ugly.

rosik:
> Please file at least an issue to luatest to have an ability to run unit tests in separate tarantool process. Currently it looks ugly.

https://github.com/tarantool/luatest/issues/122

PR #1317, opened 2021-03-15
Skip unchanged config locally

### LINKED ISSUE #1046: [4pt] Cartridge skips two-phase commit if didn't change and says "OK" when it's not

Steap2448:
The ``patch_clusterwide`` function shouldn't skip a 2pc upload if nothing changed. Instead, it'll do fair ``upload`` and ``prepare_2pc``. And the ``commit_2pc`` on every instance can then decide if the instance state is suitable for the skipping.

I didn't forget about

- [x] Tests
- [x] Changelog

Close #1046

### DIFF
--- a/cartridge/twophase.lua
+++ b/cartridge/twophase.lua
@@ -161,12 +161,21 @@ local function commit_2pc()
         return nil, err
     end
 
-
     local state = confapplier.wish_state('RolesConfigured')
 
     if state == 'Unconfigured' then
         return confapplier.boot_instance(prepared_config)
     elseif state == 'RolesConfigured' or state == 'OperationError' then
+
+        local current_config = confapplier.get_active_config()
+        if state == 'RolesConfigured' and utils.deepcmp(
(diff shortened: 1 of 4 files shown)


rosik (changes requested):
The error isn't necessary. The skipping should be preserver, just the location of decision making should be changed.

cartridge/twophase.lua
```diff
@@ -161,12 +161,21 @@ local function commit_2pc()
         return nil, err
     end
 
-
     local state = confapplier.wish_state('RolesConfigured')
 
     if state == 'Unconfigured' then
         return confapplier.boot_instance(prepared_config)
     elseif state == 'RolesConfigured' or state == 'OperationError' then
+
```
rosik, line 169: It shouldn't be skipped in case of the OperationError.

rosik (approved):

olegrok (approved):

PR #1353, opened 2021-03-29
1293 async pool map call

### LINKED ISSUE #1293: [4pt] Don't spawn fibers in pool.map_call
Spawning new fibers makes `pool.map_call` useless for the TDG - the request context isn't preserved. Instead, `map_call` should be implemented using async netbox calls.

HustonMmmavr:

I didn't forget about

- [x] Tests
- [ ] Changelog
- [ ] Documentation

Close #1293

### DIFF
--- a/cartridge/pool.lua
+++ b/cartridge/pool.lua
@@ -180,42 +180,46 @@ local function map_call(fn_name, args, opts)
 
     local retmap = {}
     local errmap = {}
-    local active_uris = {}
-    local future_map = {}
+    local future_map =  {}
     local served_uris = {}
+
+    local active_uris = {}
     local timeout = opts.timeout or 10
     local deadline = fiber.clock() + timeout
-
     do -- connect stage
         local i = 0
-        while i < #opts.uri_list and (deadline - fiber.clock()) > 0 do
+        while i < #opts.uri_list do
             i = i + 1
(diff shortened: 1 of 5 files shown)


cartridge/pool.lua
```diff
@@ -198,30 +177,100 @@ local function map_call(fn_name, args, opts)
...
+            --     - peer closed
+            --     - error was raised in remote function
+            -- - return res as array:
+            --     - res[1] - result or nil
+            --     - res[2] - error or nil (error maybe as object or string)
+            local res, err = future:wait_result(timeout) -- wait_result(0) won't yeild
+            if err ~= nil then
+                errmap[uri] = NetboxCallError:new('%q: %s', uri, err)
+            else
+                local _res, _err = unpack(res)
```
olegrok, line 254:
```lua
local clock = require('clock')

local t = {'value', 'error'}

local start = clock.time()
local a, b
for i = 1, 1e6 do
    a, b = t[1], t[2]
end
print(clock.time() - start)

local start = clock.time()
for i = 1, 1e6 do
    a, b = unpack(t)
end
print(clock.time() - start)
```

```
tarantool test.lua 
0.00028705596923828
0.026867151260376
```

So... Unpack 100 times slower than manual assign.

PR #1362, opened 2021-04-01
Preserve fiber.storage during pool.map_call

### LINKED ISSUE #1293: [4pt] Don't spawn fibers in pool.map_call

rosik:
With this patch `pool.map_call` will carefully handle `fiber.storage`. It's needed for TDG request context passing. Earlier it wasn't preserved because `pool.map_call` makes requests in separate fibers.

Initially, we were going to get rid of fibers at all. But we can't, because `netbox.connect({wait_connected = false})` doesn't work with `conn:call({is_async = true})` and throws the error: "Connection is not established, state is initial". So we still need fibers to wait until the connection is established.

Of course, we don't need `wait_connected` if it's already connected, but this optimization should be done in a separate PR.

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)

Close #1293

### DIFF
--- a/cartridge/pool.lua
+++ b/cartridge/pool.lua
@@ -138,7 +138,12 @@ local function _pack_values(maps, uri, ...)
     end
 end
 
-local function _gather_netbox_call(maps, uri, fn_name, args, opts)
+local function _gather_netbox_call(fiber_storage, maps, uri, fn_name, args, opts)
+    local self_storage = fiber.self().storage
+    for k, v in pairs(fiber_storage) do
+        self_storage[k] = v
+    end
+
     -- Perform a call, gather results and spread it across tables.
     local conn, err = connect(uri, {wait_connected = false})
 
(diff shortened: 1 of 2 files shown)


HustonMmmavr (approved):

cartridge/pool.lua
```diff
@@ -138,7 +138,13 @@ local function _pack_values(maps, uri, ...)
     end
 end
 
-local function _gather_netbox_call(maps, uri, fn_name, args, opts)
+local function _gather_netbox_call(fiber_storage, maps, uri, fn_name, args, opts)
+    local self_storage = fiber.self().storage
+    table.clear(self_storage)
```
olegrok, line 143: Actually after https://github.com/tarantool/tarantool/issues/3462 you don't need this step

rosik: Removed.

PR #1368, opened 2021-04-05
1293 async pool map call pipeline v3

HustonMmmavr:

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Close #???

### DIFF
--- a/cartridge/pool.lua
+++ b/cartridge/pool.lua
@@ -138,24 +138,16 @@ local function _pack_values(maps, uri, ...)
     end
 end
 
-local function _gather_netbox_call(fiber_storage, maps, uri, fn_name, args, opts)
+local function _gather_netbox_call(fiber_storage, maps, uri, fn_name, args, deadline)
     local self_storage = fiber.self().storage
     for k, v in pairs(fiber_storage) do
         self_storage[k] = v
     end
 
-    -- Perform a call, gather results and spread it across tables.
-    local conn, err = connect(uri, {wait_connected = false})
-
-    if conn == nil then
-        return _pack_values(maps, uri,
-            nil, err
-        )
-    else
-        return _pack_values(maps, uri,
-            errors.netbox_call(conn, fn_name, args, opts)
-        )
-    end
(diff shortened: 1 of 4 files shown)


rosik:
Not a related question, but I'm interested.

```lua

local function _pack_values(maps, uri, ...)
    -- Spread call results across tables
    for i = 1, select('#', ...) do
        local v = select(i, ...)
        maps[i] = maps[i] or {}
        maps[i][uri] = v
    end
end
```

Can it be optimized with some kind of loop unrolling? e.g.

```lua
local function _pack_values(maps, uri, ok, err, ...)
```

cartridge/pool.lua
```diff
@@ -203,29 +195,57 @@ local function map_call(fn_name, args, opts)
...
+    local deadline = fiber.clock() + timeout
+
+    for _, uri in ipairs(opts.uri_list) do
+        local conn, err = connect(uri, {wait_connected = false})
+        if conn == nil then
+            _pack_values(maps, uri, nil, err)
+        elseif conn:is_connected() then
+            local future, err = errors.netbox_call(
+                conn, fn_name, args, {is_async = true}
+            )
```
rosik, line 212: Can it ever raise an error?

cartridge/pool.lua
```diff
@@ -203,29 +195,57 @@ local function map_call(fn_name, args, opts)
...
 
-    local retmap, errmap = unpack(maps)
+    for uri, future in pairs(futures) do
+        local timeout = deadline - fiber.clock()
+        if timeout < 0 then
+            timeout = 0
+        end
+
+        _pack_values(maps, uri,
+            errors.netbox_wait_async(future, timeout, uri, fn_name)
```
rosik, line 243: Excess args 3, 4

cartridge/pool.lua
HustonMmmavr: For ex. connection became broken (but it's unlikely).
Also, I think it's better to wrap all `netbox` methods.

rosik:
Merged in #1412 instead.

Issue #660, opened 2020-03-16
Redesign membership payload dissemination


rosik:
Membership payload was designed to carry useful information, but it's size is limited by UDP packet size (1472). The more data cartridge puts there - the less free space is left for anti-entropy sync messages. So it slows down state propagation and negatively affect gossip dissemination.

We should think up another way for payload propagation.
It may be an intermediate layer between cartridge and membersip, but I believe it shouldn't be UDP.

rosik:
We've brainstormed it a little.

Membership payload is used in many places

* eventual failover (uuid, state) - every instance
* stateful failover (uuid, state) - coordinator only 
* issues (alias) - every instance
* get_topology (uuid, state, alias) - single instance
* lua-api.proxy (uuid) - single instance
* rpc.healthcheck (uuid, state) - every instance 
* topology.refine_servers_uri (uuid) - every instance
* suggestions (state) - single instance

The most tricky usecases are `rpc` and `refine_servers_uri`. They need this information on every instance. Without UDP payload dissemination it would require setting up tcp full mesh (which is expensive).

Also the `cartridge.remote-control` module doesn't support `box.session.push`.

AFAICS this issue has no solution for today.

filonenko-mikhail:
Still dead end today. Feel free to reopen kindly.

PR #1376, opened 2021-04-14
Add eventloop speedometer with refreshing membership payload

filonenko-mikhail:
In case when tarantool node use 100 cpu in tx thread
it's hard to understand is it stuck on some infinite loop or
it's just heavy operation with rarely yields

This eventloop speedometer measure event loop speed. If it is slower
than it have to be, update membership payload.

This way we can understand in membership network is member suspect
because of rarely yields on it.

I didn't forget about

- [x] Tests
- [ ] Changelog
- [ ] Documentation

Close #???


cartridge/eventloop-speedometer.lua
```diff
@@ -0,0 +1,64 @@
...
+
+    local err = (diff - tonumber64(resolution*1e9))/1e9
+
+    return err
+end
+
+local function speedometer_fiber()
+    fiber.self():name("eventloop-speedometer")
+    while true do
+        membership.set_payload('slow', nil)
```
rosik, line 27: In the worst case, this code will increment incarnation twice every 5 sec.

rosik: Also, I don't like the idea of extending payload at all, see my considerations in https://github.com/tarantool/cartridge/issues/660

rosik:
We've discussed it orally and decided to leave it unmerged for now.

My opinion: the metric is out of place.

1. If it turns out the event loop stops forever, it can't help propagate the updated payload.
2. If it only stalls temporarily, then it's all in vain.

The better place for this feature would be the metrics module. It should also track recent history, possibly as a moving average or in any other way.

PR #1386, opened 2021-04-20
Remote control loop calls

### LINKED ISSUE #1357: [4pt] Allow loop calls in remote control
Test case:

```lua
conn = require('net.box').connect('admin@localhost:3301', {wait_connected = false})
conn:eval('conn:ping()')
```

Supposing 3301 is `remoter-control` server, the code hangs forever.

Steap2448:
This patch makes a remote-control server handle every request in a separate fiber.

I didn't forget about

- [x] Tests
- [x] Changelog (unnecessary)
- [x] Documentation

Close #1357

### DIFF
--- a/cartridge/remote-control.lua
+++ b/cartridge/remote-control.lua
@@ -245,6 +245,9 @@ local function communicate(s)
         end
 
         local handler = fiber.self()
+        -- It's important to use `fiber.create` and not `fiber.new`
+        -- because it reschedules the `handler` and doesn't affect
+        -- execution order unless the async call yields.
         fiber.create(communicate_async, handler, s, sync,
             pcall, rc_eval, code, args
         )
@@ -264,6 +267,9 @@ local function communicate(s)
         end
 
         local handler = fiber.self()
+        -- It's important to use `fiber.create` and not `fiber.new`
+        -- because it reschedules the `handler` and doesn't affect
+        -- execution order unless the async call yields.
         fiber.create(communicate_async, handler, s, sync,
             rc_call, fn_name, unpack(fn_args)
         )
@@ -299,8 +305,9 @@ local function rc_handle(s)
         digest.base64_encode(salt)
     )
 
-    vars.handlers[s] = fiber.self()
-    vars.handlers[s].storage.tasks = {}
+    local handler = fiber.self()
+    handler.storage.tasks = {}
+    vars.handlers[s] = handler
     s._client_user = 'guest'
     s._client_salt = salt
 
@@ -328,7 +335,7 @@ local function rc_handle(s)
 
     s:write(greeting)
 
-    while vars.handlers[s] do
+    while vars.handlers[s] ~= nil or next(handler.storage.tasks) ~= nil do
         local ok, err = RemoteControlError:pcall(communicate, s)
         if err ~= nil then
             log.error('%s', err)
@@ -436,6 +443,7 @@ local function drop_connections()
 
     for s, handler in pairs(handlers) do
         if handler ~= fiber.self().storage.handler then
+            pcall(function() log.info('dropping %s', s) end)
             pcall(function() handler:cancel() end)
             pcall(function() s:close() end)
         end
--- a/cartridge/remote-control.lua
+++ b/cartridge/remote-control.lua
@@ -154,7 +154,7 @@ local function communicate_async(handler, s, sync, fn, ...)
 
     -- The connection was dropped and the last response was sent.
     -- Stop the communication and close the socket.
-    if vars.handlers[s] == nil and next(handler.storage.tasks) then
+    if vars.handlers[s] == nil and next(handler.storage.tasks) == nil then
         s:close() -- may raise, but nobody cares
     end
 end
--- a/cartridge/remote-control.lua
+++ b/cartridge/remote-control.lua
@@ -130,6 +130,35 @@ local function reply_err(s, sync, ecode, efmt, ...)
     s:write(mpenc_uint32(#header + #payload) .. header .. payload)
 end
 
+local function communicate_async(handler, s, sync, fn, ...)
+    local task = fiber.self()
+    task:name(handler:name())
+    task.storage.handler = handler
+    task.storage.session = handler.storage.session
+    handler.storage.tasks[task] = true
+
+    local ok, ret = fn(...)
+    if ok then
+        reply_ok(s, sync, ret)
+    elseif ffi.istype(error_t, ret) then
+        local code = ret.code
+        if code == nil and ret.errno ~= nil then
+            code = box.error.SYSTEM
+        end
+        reply_err(s, sync, code, ret.message)
+    else
+        reply_err(s, sync, box.error.PROC_LUA, tostring(ret))
+    end
+
+    handler.storage.tasks[task] = nil
+
+    -- The connection was dropped and the last response was sent.
+    -- Stop the communication and close the socket.
+    if vars.handlers[s] == nil and next(handler.storage.tasks) then
+        s:close() -- may raise, but nobody cares
+    end
+end
+
 local function communicate(s)
     local ok, buf = pcall(s.read, s, 5)
     if not ok or buf == nil or buf == '' then
@@ -215,21 +244,12 @@ local function communicate(s)
             return true
         end
 
-        local ok, ret = pcall(rc_eval, code, args)
-        if ok then
-            reply_ok(s, sync, ret)
-            return true
-        elseif ffi.istype(error_t, ret) then
-            local code = ret.code
-            if code == nil and ret.errno ~= nil then
-                code = box.error.SYSTEM
-            end
-            reply_err(s, sync, code, ret.message)
-            return true
-        else
-            reply_err(s, sync, box.error.PROC_LUA, tostring(ret))
-            return true
-        end
+        local handler = fiber.self()
+        fiber.create(communicate_async, handler, s, sync,
+            pcall, rc_eval, code, args
+        )
+
+        return true
 
     elseif iproto_code[code] == 'iproto_call' then
         local fn_name = body[0x22]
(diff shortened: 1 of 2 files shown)


cartridge/remote-control.lua
```diff
@@ -98,6 +98,24 @@ local function mpenc_uint32(n)
...
+    end
+end
+
+local function release_socket(s)
+    assert(s.ref_counter)
+    assert(s.ref_counter > 0)
+
+    s.ref_counter = s.ref_counter - 1
+    if s.ref_counter == 0 then
+        pcall(function() s:close() end)
```
olegrok, line 115:
```suggestion
        pcall(s.close, s)
```

cartridge/remote-control.lua
```diff
@@ -215,22 +234,37 @@ local function communicate(s)
...
 
-        local ok, ret = pcall(rc_eval, code, args)
-        if ok then
-            reply_ok(s, sync, ret)
-            return true
-        elseif ffi.istype(error_t, ret) then
-            local code = ret.code
-            if code == nil and ret.errno ~= nil then
-                code = box.error.SYSTEM
+        local task = fiber.new(function(...)
```
olegrok, line 237:
```suggestion
        local task = fiber.new(function(fiber_name, session)
```

cartridge/remote-control.lua
```diff
@@ -215,22 +234,37 @@ local function communicate(s)
...
+            acquire_socket(s)
+
+            local fiber_name, session = ...
```
olegrok, line 240:
```suggestion
```

cartridge/remote-control.lua
```diff
@@ -243,22 +277,36 @@ local function communicate(s)
...
 
-        local ok, ret = rc_call(fn_name, unpack(fn_args))
-        if ok then
-            reply_ok(s, sync, ret)
-            return true
-        elseif ffi.istype(error_t, ret) then
-            local code = ret.code
-            if code == nil and ret.errno ~= nil then
-                code = box.error.SYSTEM
+        local task = fiber.new(function(...)
```
olegrok, line 280:
```suggestion
        local task = fiber.new(function(name)
```

cartridge/remote-control.lua
```diff
@@ -243,22 +277,36 @@ local function communicate(s)
...
+            acquire_socket(s)
+
+            local name = ...
```
olegrok, line 283:
```suggestion
```

cartridge/remote-control.lua
```diff
@@ -421,10 +476,15 @@ local function drop_connections()
...
 
     for s, handler in pairs(handlers) do
-        if handler ~= fiber.self() then
-            pcall(function() handler:cancel() end)
-            pcall(function() s:close() end)
+        if s ~= fiber.self().storage.socket
+        and handler:status() ~= 'dead' then
+            for _, task in pairs(handler.storage.tasks) do
+                pcall(function() task:cancel() end)
+                pcall(function() release_socket(s) end)
```
olegrok, line 483:
```suggestion
                pcall(task.cancel, task)
                pcall(release_socket, s)
```

olegrok: Also asserts in `release_socket` will be ignored in this case

cartridge/remote-control.lua
```diff
@@ -421,10 +476,15 @@ local function drop_connections()
...
+            end
         end
+        pcall(function() handler:cancel() end)
+        pcall(function() release_socket(s) end)
```
olegrok, line 487:
```suggestion
        pcall(handler.cancel, handler)
        pcall(release_socket, s)
```

test/unit/remote_control_test.lua
```diff
@@ -862,3 +870,83 @@ function g.test_fd_cloexec()
...
+    t.assert_equals(conn_rc1:is_connected(), true)
+    t.assert_equals(future1:wait_result(), {true})
+
+    -- When all tasks are completed conn_rc1 is closed as well.
+    t.assert_equals(conn_rc1:is_connected(), false)
+
+    _G.storage = {}
+end
+
+function g.test_loop_call()
```
olegrok, line 914: An idea to one more test. Recursive call using single connection:

```lua
function call(n)
   if n < 10 then
       conn:call('call', n + 1)
   end
   return n
end

conn:call('call', {0})
```

cartridge/remote-control.lua
```diff
@@ -118,6 +136,7 @@ end
 local function reply_err(s, sync, ecode, efmt, ...)
     checks('?', 'number', 'number', 'string')
 
+    log.warn('reply_err')
```
olegrok, line 139: debug log

olegrok (commented):
Thanks for your patch. See several comments below (they are primary about style).

cartridge/remote-control.lua
Steap2448: All done. Thanks!

test/unit/remote_control_test.lua
```diff
@@ -154,6 +157,20 @@ function g.test_peer_uuid()
...
+
+--     local peer = conn:eval([[
+--         require('log').info(box.session.peer())
+--     ]])
+
+--     t.assert_str_matches(
+--         conn:eval('return package.loaded.fiber.name()'),
+--         'remote_control/.+:.+'
+--     )
+-- end
```
olegrok, line 172: Seems this code should be uncommented or removed

cartridge/remote-control.lua
```diff
@@ -98,6 +98,24 @@ local function mpenc_uint32(n)
...
+
+local function release_socket(s)
+    assert(s.ref_counter)
+    assert(s.ref_counter > 0)
+
+    s.ref_counter = s.ref_counter - 1
+    if s.ref_counter == 0 then
+        pcall(s.close, s)
+    end
+end
```
olegrok, line 117: LGTM but it would be great to add some comments why do you need it.

cartridge/remote-control.lua
```diff
@@ -327,6 +381,13 @@ local function rc_handle(s)
             break
         end
     end
+
+    -- Handler must wait for all tasks because socket, which is used by
+    -- them, will be closed forcefully after return. acquire_socket and
+    -- release_socket won't help.
+    for _, task in pairs(fiber.self().storage.tasks) do
+        task:join()
+    end
```
olegrok, line 390: Is it possible that new tasks will be spawned during `task:join()`?

Steap2448: Nope. New tasks are spawned only in `communicate` and `task:join` is strictly after.

cartridge/remote-control.lua
```diff
@@ -215,22 +239,36 @@ local function communicate(s)
...
-            reply_err(s, sync, code, ret.message)
-            return true
-        else
-            reply_err(s, sync, box.error.PROC_LUA, tostring(ret))
-            return true
-        end
 
+            if utils.table_count(vars.handlers) ~= 0 then
+                vars.handlers[s].storage.tasks[fiber.self():id()] = nil
+            end
```
olegrok, line 262: Why do you need this check?

First of all, if you really need to check that a table has at least one element you can call `if next(vars.handlers) ~= nil`.

But seems you need to check `vars.handlers[s] ~= nil`, so have I missed something? If yes, please add a comment.

cartridge/remote-control.lua
```diff
@@ -421,10 +482,15 @@ local function drop_connections()
     table.clear(vars.handlers)
 
     for s, handler in pairs(handlers) do
-        if handler ~= fiber.self() then
-            pcall(function() handler:cancel() end)
-            pcall(function() s:close() end)
+        if s ~= fiber.self().storage.socket
```
olegrok, line 485: When is it possible?

Steap2448:
```
handler1    handler2
   |           |______________
   |           |              |
fiber1       fiber2         fiber3
   |           |              |
 task1       task2          task3
               |
        drop_connection   
```
In such situation `fiber2` will drop all handlers and tasks started by them except those with which he shares socket.

cartridge/remote-control.lua
Steap2448: `drop_connection` applies `table.clear` on `vars.handlers`. So, I need `if next(vars.handlers) ~= nil` to check if `drop_connection` happened.

rosik:
We had significantly reworked the patch last week:

- There's a new function, `communicate_async` (executed in a separate fiber), responsible for processing an async task.
- Use the task pool as a ref counter instead of a separate one.
- Extend tests with a snippet from issue #1357 
- `drop_connections` doesn't kill any async tasks, and tests cover this case.
- The handler of the caller isn't cancelled until all requests are processed, and it's tested too.

rosik (approved):

cartridge/remote-control.lua
```diff
@@ -414,14 +432,18 @@ end
...
 -- @function drop_connections
 -- @local
 local function drop_connections()
     local handlers = table.copy(vars.handlers)
     table.clear(vars.handlers)
 
     for s, handler in pairs(handlers) do
-        if handler ~= fiber.self() then
+        if handler ~= fiber.self().storage.handler then
+            pcall(function() log.info('dropping %s', s) end)
```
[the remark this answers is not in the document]
rosik: Yes. `tostring()` fails if the socket is closed :-)

cartridge/remote-control.lua
```diff
@@ -414,14 +432,18 @@ end
...
 --- Explicitly drop all established connections.
 --
+-- Close all the sockets except the one that triggered the function.
+-- The last socket will be closed when all requests are processed.
+--
 -- @function drop_connections
 -- @local
 local function drop_connections()
     local handlers = table.copy(vars.handlers)
     table.clear(vars.handlers)
```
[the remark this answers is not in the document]
rosik: You're right. It can be rewritten without affecting the behaviour.

I usually prefer to use `table.clear` because it seems to be safer. It's nearly impossible to check (autotest) that no other code makes a closure on the table. And if it did, it could result in a very rare race which is really hard to catch. It could be something simple like

```lua
for s, handler in pairs(vars.handlers) do
    -- implicit fiber.yield()
end
```

But this is a general consideration, and it doesn't apply to this particular case. I can change it if you still want to.

test/unit/remote_control_test.lua
```diff
@@ -154,26 +154,99 @@ function g.test_peer_uuid()
...
+    rawset(_G, 'conn_2', conn_2)
+
+    -- Check that the presence of a closed connection doesn't cause errors
+    netbox.connect('superuser:3.141592@localhost:13301'):close()
 
-    t.assert_equals(conn_1:eval([[
+    -- Check that remote-control server is able to handle requests asynchronously
+    local ch = fiber.channel(0)
+    rawset(_G, 'longrunning', function() ch:put(secret) end)
+    local future = conn_2:call('longrunning', {1}, {is_async=true})
```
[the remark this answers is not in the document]
rosik: Fixed

cartridge/remote-control.lua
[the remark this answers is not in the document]
rosik: It's even covered in tests:

https://github.com/tarantool/cartridge/blob/8d05bf35f27b3ce171886472d7b616cc0d1b1ac8/test/unit/remote_control_test.lua#L196-L197

cartridge/remote-control.lua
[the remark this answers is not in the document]
olegrok: Well, I don't insist. Let's stay as is.

olegrok (approved):

PR #1424, opened 2021-05-19
Optimize `pool.map_call`

HustonMmmavr:
Optimize `pool.map_call`, with use of `net.box` futures

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

### DIFF
--- a/cartridge/pool.lua
+++ b/cartridge/pool.lua
@@ -129,22 +129,18 @@ local function connect(uri, opts)
     return conn
 end
 
-local function _gather_netbox_call(fiber_storage, retmap, errmap, uri, fn_name, args, opts)
+local function _gather_netbox_call(fiber_storage, retmap, errmap, uri, fn_name, args, deadline)
     local self_storage = fiber.self().storage
     for k, v in pairs(fiber_storage) do
         self_storage[k] = v
     end
 
-    -- Perform a call, gather results and spread it across tables.
-    local conn, err = connect(uri, {wait_connected = false})
-
-    if conn == nil then
-        errmap[uri] = err
-    else
-        local ret, err = errors.netbox_call(conn, fn_name, args, opts)
-        retmap[uri] = ret
-        errmap[uri] = err
-    end
+    local conn = connect(uri, {wait_connected = false})
+    local ret, err = errors.netbox_call(conn, fn_name, args, {
+        timeout = deadline - fiber.clock()
+    })
+    retmap[uri] = ret
+    errmap[uri] = err
 end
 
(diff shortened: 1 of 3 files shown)


rosik (approved):

cartridge/pool.lua
```diff
@@ -193,25 +197,58 @@ local function map_call(fn_name, args, opts)
...
+    for uri, future in pairs(futures) do
+        local timeout = deadline - fiber.clock()
+        if timeout < 0 then
+            timeout = 0
+        end
+
+        local ret, err = errors.netbox_wait_async(future, timeout)
+        retmap[uri] = ret
+        errmap[uri] = err
+        future:discard()
```
olegrok, line 249: IFIU discard is needed for "in-fly" requests. In case of successful (or finished with error) result it doesn't make much sense.

cartridge/pool.lua
```diff
@@ -193,25 +197,58 @@ local function map_call(fn_name, args, opts)
...
+            err = err or 'Unknown error'
+            errmap[uri] = NetboxConnectError:new('%q: %s', uri, err)
+        elseif conn:is_connected() then
+            local future, err = errors.netbox_call(
+                conn, fn_name, args, {is_async = true}
+            )
+            if future == nil then
+                errmap[uri] = err
+            end
+            futures[uri] = future
```
olegrok, line 218: [Below](https://github.com/tarantool/cartridge/pull/1424/files#diff-e337ea42ddc0b628cab1c67abd93055e8ee30bc20b5eef35f4a43acd9ae6ababR247) you use just
```lua
retmap[uri] = ret
errmap[uri] = err
```

I don't see any reasons to make code less uniform.

olegrok (commented):
Thanks for your patch. See several small comments below.

I hope a case when part of connections have been already established but another part have not been is covered. (Mix of simple calls and futures).

cartridge/pool.lua
rosik: Yes, it's a case when the `netbox_wait_async` times out.

rosik:
> I hope a case when part of connections have been already established but another part have not been is covered. (Mix of simple calls and futures).

Yes, it is. It's the reason why we use both fibers and futures. We can't do an async request unless `conn:is_connected()`. And we can't block the main fiber to wait when it's connected. That's why fresh connections start new fibers. Otherwise, it'll affect other calls and we risk catching unnecessary timeouts.

olegrok (approved):

olegrok (approved):

Issue #903, opened 2020-06-18
Fix sphinx warnings


rosik:
There are some inconsistencies with labels and ToCs

```text
/opt/cartridge/build.luarocks/build.rst/README.rst:5: WARNING: Duplicate explicit target name: "instructions".
/opt/cartridge/build.luarocks/build.rst/cartridge_api/classes/cartridge.test-helpers.cluster.rst:5: WARNING: duplicate label cartridge.test-helpers.cluster, other instance in /opt/cartridge/build.luarocks/build.rst/cartridge_api/modules/cartridge.test-helpers.rst
/opt/cartridge/build.luarocks/build.rst/cartridge_api/classes/cartridge.test-helpers.server.rst:5: WARNING: duplicate label cartridge.test-helpers.server, other instance in /opt/cartridge/build.luarocks/build.rst/cartridge_api/modules/cartridge.test-helpers.rst
/opt/cartridge/build.luarocks/build.rst/cartridge_api/modules/cartridge.rst:5: WARNING: duplicate label cartridge, other instance in /opt/cartridge/build.luarocks/build.rst/index.rst
/opt/cartridge/build.luarocks/build.rst/cartridge_api/modules/cartridge.test-helpers.rst:22: WARNING: duplicate label cartridge.test-helpers.server, other instance in /opt/cartridge/build.luarocks/build.rst/cartridge_api/classes/cartridge.test-helpers.server.rst
/opt/cartridge/build.luarocks/build.rst/cartridge_api/modules/cartridge.test-helpers.rst:37: WARNING: duplicate label cartridge.test-helpers.cluster, other instance in /opt/cartridge/build.luarocks/build.rst/cartridge_api/classes/cartridge.test-helpers.cluster.rst
/opt/cartridge/build.luarocks/build.rst/cartridge_dev.rst:5: WARNING: Duplicate explicit target name: "install".
/opt/cartridge/build.luarocks/build.rst/cartridge_dev.rst:5: WARNING: Duplicate explicit target name: "install".
/opt/cartridge/build.luarocks/build.rst/cartridge_dev.rst:5: WARNING: Duplicate explicit target name: "install".
/opt/cartridge/build.luarocks/build.rst/index.rst:13: WARNING: toctree contains reference to nonexisting document 'cartridge_cli/index'
[...]

PR #1430, opened 2021-05-27
Move rst substitutions from documents to rst_epilog variable

NickVolynkin:
Removes one more warning from the build logs
Part of #903


NickVolynkin:
@rosik I've canceled all workflows as there are no code changes.

rosik (approved):

Issue #1085, opened 2020-10-07
[4pt] schema.yml disappears from page "Code"


dsamirov:
I create file named "schema.yml" on the page `/admin/cluster/code`. Then I press Apply and this file disappears.

However downloaded config contains my uploaded data.

rosik:
Proposal: get rid of the "schema" page. 

It's almost useless in comparison to the "code" page.

There are two features missing on the "code" page:

- The "validate" button. But we can add it.
- Schema example. Can be added similar to https://github.com/tarantool/cartridge/pull/1194

This would also resolve https://github.com/tarantool/cartridge/issues/1082, https://github.com/tarantool/cartridge/issues/1201, https://github.com/tarantool/cartridge/issues/1203

rosik:
N.B. Originally the issue is caused by the following line in frontend code:

```
src/store/reducers/files.reducer.js:const ignoreFiles = ['schema.yml']
```

PR #1438, opened 2021-06-07
New validate_config method

Steap2448:

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Related to #1085


cartridge/clusterwide-config.lua
```diff
@@ -249,6 +250,37 @@ clusterwide_config_mt = {
             return self._checksum
         end,
 
+        new = function(self, patch)
```
olegrok, line 253: I suggest to rename it to "patch". But I like this change - we need it in TDG code.

cartridge/clusterwide-config.lua
```diff
@@ -249,6 +250,37 @@ clusterwide_config_mt = {
...
+            checks('ClusterwideConfig', 'table')
+
+            local new_config = self:copy()
+            for k, v in pairs(patch) do
+                if patch[k] ~= nil and patch[k .. '.yml'] ~= nil then
+                    local err = NewConfigError:new(
+                        'Ambiguous sections %q and %q',
+                        k, k .. '.yml'
+                    )
+                    log.error('%s', err)
```
olegrok, line 263: I think it could be moved out of this function. Such simple function shouldn't produce any side-effects.

cartridge/twophase.lua
```diff
@@ -507,29 +507,9 @@ local function _clusterwide(patch)
...
-            clusterwide_config_new:set_plaintext(k, v)
-        else
-            if not string.endswith(k, '.yml') then
-                clusterwide_config_new:set_plaintext(k, box.NULL)
-                k = k .. '.yml'
-            end
-
-            clusterwide_config_new:set_plaintext(k, yaml.encode(v))
-        end
+    local clusterwide_config_new, err = clusterwide_config_old:patch_copy(patch)
```
rosik, line 510: I suggest keeping it split as before:

```lua
local clusterwide_config_new = clusterwide_config_old:copy()
local _, err = clusterwide_config_new:patch(patch)
```

But +1 for moving it to clusterwide module, I like it too.

cartridge/webui/api-config.lua
```diff
@@ -252,6 +266,66 @@ local function set_sections(_, args)
...
+        local ddl_manager = assert(service_registry.get('ddl-manager'))
+        local ok, err = ddl_manager.check_schema_yaml(args.as_yaml)
+        if not ok then
+            if err.class_name == ddl_manager.CheckSchemaError.name then
+                return { error = err.err }
+            else
+                return nil, err
+            end
+        end
+    end
```
rosik, line 296: Please, leave a comment on why it's necessary. The schema is already checked as a part of `confapplier.validate_config`, but it's skipped on non-leader instances. That's why we use `ddl_manager.check_schema_yaml` (which uses rpc)

rosik: It also could be optimized with an `if not is_leader ...`

cartridge/twophase.lua
olegrok: > I suggest keeping it split as before:

In case if we consider clusterwide_config as immutable object correct way is to create copy for each object modification.

Steap2448: I would prefer to treat it as an immutable object as well.

cartridge/clusterwide-config.lua
Steap2448: Is "patch_copy" ok? What do you think?

cartridge/clusterwide-config.lua
Steap2448: Agreed!

cartridge/webui/api-config.lua
Steap2448: Done!

cartridge/clusterwide-config.lua
olegrok: Returning back to https://github.com/tarantool/cartridge/pull/1438#discussion_r647789718 I don't understand why "copy" word is needed. But up to you.

cartridge/twophase.lua
rosik: Ok

rosik (approved):

PR #1446, opened 2021-06-21
fix unhandled error on config load

olegrok:
Before this patch if we tried to load yaml file that couldn't be
parsed into Lua table we got following error:

```
.../.rocks/share/tarantool/cartridge/clusterwide-config.lua:320:
bad argument #1 to 'pairs' (table expected, got string)
```

This patch fixes such case and add a check that parsed config
is a table.


rosik (approved):
Thank you for the patch, LGTM.

PR #1452, opened 2021-06-29
Add failover polling

### LINKED ISSUE #1364: [2pt] When failover settings changed not from UI need to reload UI page to get them

aShelist:
Close #1364


rosik:
Corner case:

1. Open Cartridge dashboard in two browser tabs,
2. Open the failover dialog in the first tab.
3. In the second tab edit the failover params (e.g. enable stateful failover)
4. Return back to the first tab. It still should stow the same disabled state, but in the background failover button already shows new state 
image

Where the problem starts:

5. In the first tab close the dialog and open it again.
6. It should already show updated params, but it still shows "disabled" radio-button selected.

My wish: fetch failover params (with a spinner animation) when opening the failover dialog.

aShelist:
> My wish: fetch failover params (with a spinner animation) when opening the failover dialog.

Done

rosik (approved):

PR #1491, opened 2021-07-21
Download config file on cartridge

OlgaYakovleva:
Install download file module
Add setups to use download file module
Add test to download config file


OlgaYakovleva:
Some problems with CI, closing

PR #1490, opened 2021-07-21
Added membership and vshard tabs to server details

yngvar-antonsson:
Introduce 3 new tabs in server details dialog:

## membership
image

## vshard-router
image

## vshard-storage
image

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] ~~Documentation~~

Close #1196

PR #1516, opened 2021-08-06
Fix flaky multisharding_two test

rosik:
The servers comparison used to fail sometimes because it could contain
another unconfigured spare server. Fix it by ensuring the discovery
finishes.

Also, extend the test with vshard_storage info.

The test was introduced in #1490 

I didn't forget about

- [x] Tests
- [x] ~~Changelog~~
- [x] ~~Documentation~~

### DIFF
--- a/test/integration/multisharding_two_test.lua
+++ b/test/integration/multisharding_two_test.lua
@@ -62,7 +62,17 @@ g.before_all = function()
 
     g.server:start()
     t.helpers.retrying({}, function()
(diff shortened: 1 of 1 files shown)


yngvar-antonsson:
Flaky test in CI:
https://github.com/tarantool/cartridge/runs/3260279593?check_suite_focus=true

yngvar-antonsson (approved):

PR #1528, opened 2021-08-16
Add vshard issues

### LINKED ISSUE #941: [2pt] Indicate vshard issues
Inspired by #875: such problems could be monitored with cartridge [issues feature](https://github.com/tarantool/cartridge/issues/289)

See also #1112

yngvar-antonsson:
Introduce two new issues, forward them from vshard:

```
warning: VShard alert on localhost:13305 (router): 1000 buckets are not discovered
```

```
warning: VShard alert on localhost:13305 (router): Invalid configuration:
  probably router's cfg.bucket_count is different from storages' one, difference is 500"
```

Also, fix `api_query_test` which started to fail more intensively because of the discovery issue.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #941

### DIFF
--- a/cartridge/issues.lua
+++ b/cartridge/issues.lua
@@ -355,6 +355,35 @@ local function list_on_instance(opts)
         end
     end
 
+
+    -- Vshard issues
+    local vshard = package.loaded.vshard
+
+    if vshard then
+        local routers = vshard and vshard.router.internal.routers or {}
+        if next(routers) ~= nil then
+            for _, router in pairs(routers) do
+                local alerts = router:info().alerts
+                for _, alert in ipairs(alerts) do
+                    if alert[1] == 'UNKNOWN_BUCKETS'
+                    or alert[1] == 'INVALID_CFG'
+                    then
+                        local issue = {
+                            level = 'warning',
+                            topic = 'vshard',
+                            replicaset_uuid = replicaset_uuid,
+                            instance_uuid = instance_uuid,
+                            message = alert[2]
+                        }
+                        table.insert(ret, issue)
+                    end
+                end
+            end
+        end
+    end
+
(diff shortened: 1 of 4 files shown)


cartridge/issues.lua
```diff
@@ -344,6 +349,44 @@ local function list_on_instance(opts)
...
+        if routers == nil or next(routers) == nil then
+            goto continue
+        end
+
+        for _, router in pairs(routers) do
+            local alerts = router:info().alerts
+            for _, alert in ipairs(alerts) do
+                local type, message = alert[1], alert[2]
+                log.info(require('yaml').encode(alert))
+                if type == 'UNKNOWN_BUCKETS'
```
rosik, line 369: With this predicate, a cluster shows issues even before the vshard is bootstrapped, and it's no good, I think. Also, WebUI tests fail because of it.

rosik (changes requested):

rosik:
We've discussed this patch with @no1seman and found it to be harmful.

1. During ansible deployments, it's usual to check for the absence of issues. 
1. There're projects with 200K buckets. 

If we merge this patch, every ansible step will take up to 200 seconds which is unacceptable.

An alternative solution could be introducing a new issues level, but then it wouldn't be an issue at all.

As for the "Invalid configuration" issue, I think it doesn't have much value either. Buckets count difference seems to be nearly impossible to me.

I suggest that we admit this experiment has failed. There's nothing to monitor in the current version of vshard. And if something emerges there someday, this patch could be resurrected.

no1seman:
OK, let it be closed but there is solution to make special severirty for such issues and this will solve the problem. May be in some unknown future....

Issue #1498, opened 2021-07-29
[RFC] Custom messages in `cartridge.issues` UI


opomuc:
Sometimes a developer wants to throw an error/warning so the user will see it right away. 

For example when configuring connection to Kafka (or other storage) one would need to apply config with connection parameters. This is an asynchronous action. If no connection is established, user won't know it right away without looking into logs.

I propose something along the lines:
```
local issues = require('cartridge.issues')

issues.warn('A warning message')
```

As for the issue structure:
```
issues.warn(msg, ...)
...
->
        local issue = {
            level = 'warning',
            topic = 'user',
            replicaset_uuid = cartridge.confapplier.get_replicaset_uuid(),
            instance_uuid = cartridge.confapplier.get_instance_uuid(),
            message = string.format(msg, ...)
        }
```
Same with other log levels: `error`, `critical`

PR #1532, opened 2021-08-16
Custom issues API rfc

yngvar-antonsson:
It's an API RFC for #1498


rosik (commented):
I think it should be implemented in a different way. Let's introduce a new role callback - `get_issues()`. And a role would return a set of issues, just like the auth backend works.

cartridge/issues.lua
```diff
@@ -355,6 +361,27 @@ local function list_on_instance(opts)
         end
     end
 
+    -- add custom issues from each role
+    local registry = require('cartridge.service-registry')
+    for role_name, role in pairs(registry.list()) do
+        if role.get_issues ~= nil then
+            for _, issue in ipairs(role.get_issues()) do
```
rosik, line 368: Check it's a function and pcall it. See a similar code in `roles.lua`:

https://github.com/tarantool/cartridge/blob/master/cartridge/roles.lua#L342-L351

cartridge/issues.lua
```diff
@@ -355,6 +361,27 @@ local function list_on_instance(opts)
...
+                    table.insert(ret, {
+                        level = issue.level,
+                        message = issue.message,
+                        topic = issue.topic or 'custom',
+                        instance_uuid = instance_uuid,
+                        replicaset_uuid = replicaset_uuid,
+                    })
+                else
+                    log.error(('Got issue with wrong format from role %s: level = %s, topic = %s, message = %s'):
+                        format(role_name, tostring(issue.level), tostring(issue.topic), tostring(issue.message)))
```
rosik, line 379: It could be an issue too

cartridge/issues.lua
```diff
@@ -91,6 +91,12 @@ local limits_ranges = {
     clock_delta_threshold_warning = {0, math.huge},
 }
 
+local allowed_levels = {
+    critical = true,
+    warning = true,
+    info = true,
+}
+
```
rosik, line 99: I think we could go without this restriction. If a user wants to define a custom level, why would we forbid him

test/integration/issues_test.lua
```diff
@@ -396,3 +397,48 @@ g.after_test('test_aliens', function()
...
+                instance_uuid = g.master.instance_uuid,
+                replicaset_uuid = g.master.replicaset_uuid,
+            }
+        })
+    end)
+
+    g.master.net_box:eval([[
+        local cartridge = require('cartridge')
+        cartridge.service_set('custom_role_with_issues', nil)
+    ]])
```
rosik, line 443: I suggest moving it to `g.after_test('test_custom_issues', function() ... end)`

test/integration/issues_test.lua
```diff
@@ -396,3 +397,48 @@ g.after_test('test_aliens', function()
...
+                    {
+                        level = 'critical',
+                        topic = nil,
+                        message = 'critical message',
+                    },
+                }
+            end,
+        })
+    ]])
+    t.helpers.retrying({}, function()
```
rosik, line 421: Is retrying necessary here?

test/integration/issues_test.lua
```diff
@@ -396,3 +397,48 @@ g.after_test('test_aliens', function()
...
+        t.assert_equals(helpers.list_cluster_issues(g.master), {
```
rosik, line 422: Please, add one more test with `get_issues` raising an error.

cartridge/issues.lua
```diff
@@ -50,6 +50,21 @@
...
+-- Custom issues (defined by user):
+-- Collected from role callback `get_issues`. It should return table with
+-- issues parameters:
+-- ```lua
+-- function get_issues()
+--     return {
+--         {
+--             level = 'custom-level',
+--             topic = 'custom-topic',
+--             message = 'custom message',
```
rosik, line 62:
```suggestion
--             message = 'custom message',
--             -- all mentioned fields are optional,
--             -- other fields are ignored
```

cartridge/issues.lua
```diff
@@ -355,6 +370,25 @@ local function list_on_instance(opts)
...
+    local registry = require('cartridge.service-registry')
+    for _, role in pairs(registry.list()) do
+        if type(role.get_issues) == 'function' then
+            local ok, custom_issues = pcall(role.get_issues)
+            if ok then
+                for _, issue in ipairs(custom_issues) do
+                    table.insert(ret, {
+                        level = issue.level or '',
+                        message = issue.message or '',
+                        topic = issue.topic or 'custom',
```
rosik, line 383:
```suggestion
                        topic = issue.topic or role_name,
```

cartridge/issues.lua
```diff
@@ -355,6 +370,25 @@ local function list_on_instance(opts)
         end
     end
 
+    -- add custom issues from each role
+    local registry = require('cartridge.service-registry')
+    for _, role in pairs(registry.list()) do
+        if type(role.get_issues) == 'function' then
+            local ok, custom_issues = pcall(role.get_issues)
+            if ok then
```
rosik, line 378: Else add one more issue:

```lua
level = warning
topic = role_name
message = err
```

cartridge/issues.lua
```diff
@@ -355,6 +370,25 @@ local function list_on_instance(opts)
...
+                for _, issue in ipairs(custom_issues) do
```
rosik, line 379: Check if custom_issues is actually a table. Otherwise it's an issue too.

test/integration/issues_test.lua
```diff
@@ -396,3 +397,65 @@ g.after_test('test_aliens', function()
...
+        },
+        {
+            level = '',
+            topic = 'custom',
+            message = '',
+            instance_uuid = g.master.instance_uuid,
+            replicaset_uuid = g.master.replicaset_uuid,
+        }
+        -- since custom_role_with_issues_error.get_issues() returns an error,
+        -- there will be no issues from that role in list
```
rosik, line 451: That's no good. There should be one

cartridge/issues.lua
```diff
@@ -355,6 +372,37 @@ local function list_on_instance(opts)
...
+                    }}
+                else
+                    custom_issues = {{
+                        level = 'warning',
+                        message = tostring(custom_issues),
+                    }}
+                end
+            end
+
+            for _, issue in ipairs(custom_issues) do
```
rosik, line 394: What if `custom_issues` isn't a table

cartridge/issues.lua
```diff
@@ -355,6 +390,24 @@ local function list_on_instance(opts)
         end
     end
 
+    -- add custom issues from each role
+    local registry = require('cartridge.service-registry')
+    for role_name, M in pairs(registry.list()) do
```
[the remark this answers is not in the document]
rosik: Extended changelog.

rosik (approved):

PR #1535, opened 2021-08-17
Make sure no error messages in Tarantool log after testing

ligurio:
**NOTE**: It's a WIP implementation. I want to find errors in Tarantool logs on teardown. This patch is an MVP for the demonstration of the idea.

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Close #???

### DIFF
--- a/cartridge/test-helpers/cluster.lua
+++ b/cartridge/test-helpers/cluster.lua
@@ -23,6 +23,8 @@ local Cluster = {
     failover = 'disabled',
 }
 
+DEFAULT_LOG_PATTERN = '>E'
+
 function Cluster:inherit(object)
     setmetatable(object, self)
     self.__index = self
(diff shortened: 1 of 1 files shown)


cartridge/test-helpers/cluster.lua
```diff
@@ -103,6 +107,8 @@ function Cluster:initialize()
             },
             env = {
                 TARANTOOL_PASSWORD = self.cookie,
+                TARANTOOL_LOG = fio.pathjoin(workdir,
+                    string.format('tarantool-%s.log', self.alias)),
```
rosik, line 111: With this change, you unconditionally deny `luatest -c` debugging.

rosik (commented):
There're lots of tests that intentionally cause errors. Splitting then would be rather difficult.

But now they fail due to the rough globals and the strict mode.

I don't think this minimal Viable product is ever viable.

rosik:
Obsolete. Closing

PR #1547, opened 2021-08-31
Support cartridge_hide_all_rw in tarantool_variables

### LINKED ISSUE #1079: [1pt] Allow to hide "Make all instances writeable"
Some systems could work only when master is single instance. Usage of "all_rw" option could be easily banned with topology check. However checkbox still will be shown in WebUI. So it could confuse users.

Therefore I suggest to implement a way to hide "Make all instances writeable" option.

usenko-timur:
Variables:

* `cartridge_hide_all_rw` - hide 'Make all instances writeable' configuration field 

Example for browser runtime:

```js
if (!window.__tarantool_variables) window.__tarantool_variables = {};
window.__tarantool_variables.cartridge_hide_all_rw = true;
```

Example to fine-tune in from backend:

```lua
local frontend = require('frontend-core')
frontend.set_variable('cartridge_hide_all_rw', true)
```

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1079

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,9 @@ and this project adheres to
 [Unreleased]
 -------------------------------------------------------------------------------
 
+- 'Make all instances writeable' configuration field can be hidden via
+  frontend-core's ``set_variable`` feature or at runtime.
+
 -------------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


rosik (approved):
I've edited the test. Please, re-check it.

Issue #1201, opened 2020-12-23
[2pt] Ambiguous notifications in WebUI shema page


rosik:
![image](https://user-images.githubusercontent.com/2205188/102988362-dd5e0e00-4524-11eb-85af-8985bea7042c.png)

## Steps to reproduce

1. Open schema page in WebUI;
2. Spoil the schema, e.g. `spaces: false`;
3. Click the "validate" button;
4. Both green and red notifications are shown at the same time.

rosik:
Resolved in #1492

PR #1566, opened 2021-10-01
Cartridge setup snapshot

OlgaYakovleva:
Setup cypress-image-snapshots on cartridge
Here is good areticle about this plugin
https://medium.com/norwich-node-user-group/visual-regression-testing-with-cypress-io-and-cypress-image-snapshot-99c520ccc595

Every page with the minimal setup: dashboard, users, config, code
Created test with snapshots for #1201 in code-page.spec.js

Added in commands/index.js command setResolution to allow set standard and non-standart resolutions for a page before snapshot

set parameters for snapshots in addMatchImageSnapshotCommand


OlgaYakovleva:
This is closing as updated version (because of cartridge redesign) has been create

PR #1567, opened 2021-10-04
Relocate default admin

### LINKED ISSUE #1316: Allow to disable default admin user
Cartridge allows to specify set of callbacks to create custom auth backend. It's ok and it's useful. However, at the same time cartridge checks that login=admin and password=cluster-cookie could be provided as well.

This unexpected feature could cause conflicts with custom auth backend. I suggest to add a way/option to disable any default cartridge users.

Steap2448:
Move the default admin to the ``auth-backend`` module. It can be disabled with a ``TARANTOOL_ADMIN_DISABLED`` env variable or a ``.custom.admin_disable()`` method.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] ~~Documentation~~

Close #1316


cartridge.lua
```diff
@@ -593,12 +593,14 @@ local function cfg(opts, box_opts)
...
+    elseif type(auth_backend) ~= 'table' then
+        return nil, CartridgeCfgError:new(
+            "Auth backend must export a table, got %s",
+            type(auth_backend)
+        )
     end
 
-    local ok, err = CartridgeCfgError:pcall(function()
-        local ok = auth.set_callbacks(auth_backend)
-        return ok
```
rosik, line 601: Historical remark. One could ask: why do we need these `local ok, return ok`, why not simply call `return set_callbacks()`? The reason was in the `checks` which spoiled the message. Simple `return` drains a frame from the call stack, leaving `pcall` on the top. And the error looks as follows:

```log
bad argument callback.check_password to nil
```

This patch eliminates the generation of error messages based on `checks`, so we can rewrite this code in a more straightforward way.

cartridge/auth-backend.lua
```diff
@@ -97,6 +138,15 @@ end
 local function add_user(username, password, fullname, email)
     checks('string', 'string', '?string', '?string')
 
+    if username == ADMIN_USER.username
+    and is_builtin_admin_enabled()
```
Steap2448, line 142: Why do you discard `get_admin_username` function?

rosik: I think `get_admin_username` name is misleading. It implies one can `set_admin_username`, but that's not true. In fact, `get_admin_username` is a twisted `is_enabled`. It's the Occam's razor: entities should not be multiplied beyond necessity.

Steap2448: I don't insist on this particular name, it can be renamed. 

> entities should not be multiplied beyond necessity.
As a result, there are code duplications

rosik (approved):

cartridge/auth-backend.lua
olegrok: How following case will be handled?
  - Disable admin user
  - Add custom admin
  - Enable admin user

cartridge/auth-backend.lua
```diff
@@ -86,6 +120,13 @@ end
 local function get_user(username)
     checks('string')
 
+    if ADMIN_USER.enabled and username == ADMIN_USER.username then
+        return {
+            username = ADMIN_USER.username,
+            fullname = ADMIN_USER.fullname,
+        }
```
olegrok, line 127: nit: I guess it could be moved into separate function. Such table is used in several places.

PR #1573, opened 2021-10-14
ci: add reusable testing workflow

ylobankov:
The idea of this workflow is to be a part of the `integration.yml`
workflow from the tarantool repo to verify the integration of the
cartridge module with an arbitrary tarantool version.

This workflow is not triggered on a push to the repo and cannot be run
manually since it has only the `workflow_call` trigger. This workflow
will be included in the tarantool development cycle and called by the
`integration.yml` workflow from the tarantool project on a push to the
master and release branches for verifying integration of tarantool with
cartridge.

Part of tarantool/tarantool#6512
Part of tarantool/tarantool#5265
Part of tarantool/tarantool#6056

### DIFF
--- a/.github/workflows/reusable-backend-test.yml
+++ b/.github/workflows/reusable-backend-test.yml
@@ -1,0 +1,86 @@
+name: Reusable Backend Test
+
+on:
+  workflow_call:
+    inputs:
+      artifact_name:
+        description: 'The name of the tarantool build artifact'
+        default: ubuntu-focal
+        required: false
+        type: string
+
+env:
+  # Skip building frontend in `tarantoolctl rocks make`.
+  CMAKE_DUMMY_WEBUI: true
+  # Prerequisite for some etcd-related tests.
+  ETCD_PATH: etcd-v2.3.8/etcd
+
+jobs:
+  run_tests:
+    runs-on: ubuntu-20.04
+    steps:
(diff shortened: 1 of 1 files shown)


ylobankov:
Related PR https://github.com/tarantool/tarantool/pull/6513.

.github/workflows/reusable-backend-test.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      # Stop Mono server. This server starts and listen to 8084 port that is
+      # used for tests. Relevant to GitHub-hosted runners.
+      - name: 'Stop Mono server'
+        run: sudo kill -9 $(sudo lsof -t -i tcp:8084)
+        continue-on-error: true
+
+      # Run tests
+      - name: 'Run luatest'
+        run: .rocks/bin/luatest -v -b
```
VitaliyaIoffe, line 73: It can bring problems if we have not executed the next `tarantoolctl rocks remove luatest` on ours runners.

ylobankov:
> I have some questions about versions. As I remember you said that you had got versions from cartridge workflows. So, I'd like to know when devs update versions and how we would update versions in this workflow. Manually? Can we take the last versions?

Yes, we can take the latest version to not update the luatest version manually. Fixed.

Steap2448 (approved):

PR #1577, opened 2021-10-15
Don't validate config on disabled roles

### LINKED ISSUE #859: Cartridge shouldn't call validate_config if role disabled
Subj: https://github.com/tarantool/cartridge/blob/master/cartridge/roles.lua#L256

Configuration is checked even for disabled roles.
In contrast apply_config (that's correct) is called only for enabled roles (https://github.com/tarantool/cartridge/blob/master/cartridge/roles.lua#L295).

Steap2448:
`validate_config` is not called on disabled roles. It doesn't affect dependencies (both hidden and not) and permanent roles. 

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #859


cartridge/roles.lua
```diff
@@ -299,8 +299,22 @@ local function validate_config(conf_new, conf_old)
         error(err, 2)
     end
 
+    local disabled_roles = {}
+    local state = require('cartridge.confapplier').get_state()
+    if state == 'RolesConfigured' or state == 'OperationError' then
```
yngvar-antonsson, line 304: If there any tests for OperationError state?

test/integration/config_test.lua
```diff
@@ -245,6 +245,15 @@ function g.test_upload_fail()
     ]])
 end
 
+g.before_test('test_validate', function()
```
yngvar-antonsson, line 248: And please add a commit where that test fails without new fixes.

cartridge/roles.lua
```diff
@@ -299,8 +299,22 @@ local function validate_config(conf_new, conf_old)
         error(err, 2)
     end
 
+    local disabled_roles = {}
```
olegrok, line 302: Please, add a comment why do you check "disabled" roles, not just "enabled".

Is it related anyhow to permanent/hidden roles? Validate config should not be skipped for them.

test/integration/config_test.lua
olegrok: Or I suggest to extend error message: "Validation should not be called because role is disabled"

test/integration/config_test.lua
```diff
@@ -245,46 +245,102 @@ function g.test_upload_fail()
...
+        return require('cartridge.confapplier').get_state()
+    ]])
+    t.assert_equals(state, 'RolesConfigured')
+
+    g.cluster.main_server:eval([[
+        local mymodule = package.loaded['mymodule-permanent']
+        mymodule.apply_config = _G._apply_config
+        _G._apply_config = nil
+
+        hidden = package.loaded['mymodule-hidden']
```
olegrok, line 339: This variable is not used anyhow

olegrok (approved):

PR #1588, opened 2021-10-29
Check config checksum mismatch during ``rpc.get_candidates``

### LINKED ISSUE #1575: Fix `rpc.get_candidates` function
```lua
log.info(err)
log.info(candidates)
```

- insert code:

  ```lua
      fiber.sleep(3)
  end
  ```
  
Steap2448:
New rpc option ``same_topology`` to filter replicas with topology checksum mismatch. Checksum to compare is obtained from membership.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1575

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,10 @@ Changed
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 - Update ``http`` dependency to 1.1.1
+- ``config_checksum`` is added to membership payload. It is used to detect
+  config mismatch.
+- Members with config checksum mismatch are not returned as a result of
+  ``cartridge.rpc.get_candidates`` when ``opts.healthy_only`` is ``true``.
 
 -------------------------------------------------------------------------------
 [2.7.3] - 2021-10-27
--- a/cartridge/clusterwide-config.lua
+++ b/cartridge/clusterwide-config.lua
@@ -239,9 +239,15 @@ clusterwide_config_mt = {
             end
         end,
 
-        get_checksum = function(self)
-            checks('ClusterwideConfig')
+        get_checksum = function(self, section_name)
+            checks('ClusterwideConfig', '?string')
             assert(self._plaintext ~= nil)
+
+            if section_name then
+                local checksum = digest.crc32.new()
+                checksum:update(self._plaintext[section_name] or '')
+                return checksum:result()
+            end
 
             if self._checksum == nil then
                 self:generate_checksum()
--- a/cartridge/confapplier.lua
+++ b/cartridge/confapplier.lua
@@ -252,6 +252,10 @@ local function apply_config(clusterwide_config)
     )
 
     vars.clusterwide_config = clusterwide_config
+    membership.set_payload(
+        'topology_checksum',
+        vars.clusterwide_config:get_checksum('topology.yml')
+    )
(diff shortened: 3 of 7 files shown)


CHANGELOG.rst
```diff
@@ -23,6 +23,9 @@ Changed
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 - Update ``http`` dependency to 1.1.1
+- ``config_checksum`` is added to membership payload.
```
olegrok, line 26: I see a some failures in TDG test suite after apply this changes. So, I don't see any motivation and problem description why this changes were done. Please describe.

olegrok: TDG test report:
== 37 failed, 875 passed, 19 skipped, 1 xfailed, 2 xpassed, 94 errors in 1460.53s (0:24:20) ===

Steap2448: The original problem was `get_candidates` (called on `A1`) returning instances not suitable for requested role (`B1`). For example it occurs when twophase on `B1` is stuck on commit phase and new role has not been applied on it. `A1` has no idea that `B1` hasn't applied new config and rightfully thinks that `B1` is a good candidate for a role, that was previously assigned to it, judging from its local topology. The problem is in different applied configs.

So, we can provide `get_candidates` with information about config checksum on different instances. In such way there is a guarantee that conclusions made from the local config are also true for all other instances that have the same config checksum. It is okay to use `membership` for such purpose since `get_candidates` already uses membership's payload to get instances' states. For instances whose checksum is different there are no reliable assumptions, so they are treated as unhealthy.

However, there is a easy way around, which is to say that `get_candidates` returns **candidates**, so there is no guarantee that there are indeed requested roles applied on them.

olegrok: First of all, patch shouldn't break backward compatibility. Please, verify your changes agains TDG. 
Secondly, probably it's not appropriate place for membership usage. We have one "race" during apply/validate config and eventual membership can't solve it.

One solution is to ban get_candidates during apply_config and validate_config. But it could break backward compatibility.

olegrok: So. I realized that we shouldn't use config checksum anyway. rpc_candidates is based on topology. So... We should use "topology" checksum, not whole config checksum.

Steap2448: > One solution is to ban get_candidates during apply_config and validate_config

It seems that the problem remains because it requires a reliable way to know what is happening on remote replica. 

Backward compatibility can be provided by introducing additional parameter to `rpc.call`/`rpc.get_candidates` that will be disabled by default.

Steap2448: >We should use "topology" checksum, not whole config checksum

Agreed. However tools to get both checksum from config are implemented.

olegrok: > Backward compatibility can be provided by introducing additional parameter to rpc.call/rpc.get_candidates that will be disabled by default.

It's a bit inappropriate. We don't need new options disabled by default. At first it will be unused. Secondly, I don't want to add new argument to all existing rpc.* calls.

Could we perform some additional request to an instance who definitely knows topology?

Steap2448: > Could we perform some additional request to an instance who definitely knows topology?

It would be an external state provider which we don't have.

olegrok: Well, I see.

So I think single way to solve this issue is to make user to use topology from applied configuration and specify uri manually. Seems we can't propose reliable way to know roles/topology during appy_config.
I propose to close an issue with this recipe.

PR #1597, opened 2021-11-08
Fix cartridge.cfg params description

### LINKED ISSUE #1596: cartridge.cfg params aren't described enough
- `bucket_count` could be set only once, before first cluster start
- `vshard_groups` could be set with two notations:

```lua
vshard_groups = {
        -- both notations are valid
        ['cold'] = {bucket_count = 2000},
        'hot',
    },
```

yngvar-antonsson:
I didn't forget about

- [x] Documentation

Close #1596

### DIFF
--- a/cartridge.lua
+++ b/cartridge.lua
@@ -159,7 +159,7 @@ end
 --
 -- @tparam ?number opts.bucket_count
 --  bucket count for vshard cluster. See vshard doc for more details.
---  Could be set only **once**, before first cartridge start and coundn't be
+--  Can be set only **once**, before the first run of Cartridge application, and can't be
 --  changed after that.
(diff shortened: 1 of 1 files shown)


cartridge.lua
```diff
@@ -159,12 +159,17 @@ end
 --
 -- @tparam ?number opts.bucket_count
 --  bucket count for vshard cluster. See vshard doc for more details.
+--  Could be set only **once**, before first cartridge start and coundn't be
```
NickVolynkin, line 162: Why in the past tense? 
```suggestion
--  Can be set only **once**, before the first run of Cartridge application, and can't be
```

NickVolynkin (approved):

PR #1613, opened 2021-11-18
allow to extract filename from http request body

olegrok:
This patch introduces the third argument for metadata of http
request parsing (the second is reserved for an error). Before this
patch function http_read_body returned only file content. However
some of our customers want to know filename as well. It's not so
hard to implement because body headers have such info and the main
goal of this patch is to implement needed logic for header
extraction.

---
Need for https://github.com/tarantool/tdg2/issues/1207

### DIFF
--- a/cartridge/utils.lua
+++ b/cartridge/utils.lua
@@ -337,6 +337,20 @@ local function is_email_valid(str)
     return true
 end
 
+--- Read HTTP request body from request object.
+-- There are two cases:
+--   * Text body
+--   * File body
+-- In the first case this function returns text as is.
+-- In the second case file content should be additionally
+-- processed and file itself should be separated from
+-- its headers.
+-- @function http_read_body
+-- @local
+-- @tparam table req A HTTP request object.
+-- @treturn[1] table The same table `tbl`.
+-- @treturn[1] nil.
+-- @treturn[1] table File properties (e.g. filename).
 local function http_read_body(req)
     local req_body = req:read()
     local content_type = req.headers['content-type'] or ''
@@ -356,13 +370,21 @@ local function http_read_body(req)
     -- that have a special meaning with % to escape them.
     -- A list of special characters is ().+-*?[]^$%
     local boundary_line = string.gsub('--'..boundary, "[%(%)%.%+%-%*%?%[%]%^%$%%]", "%%%1")
-    local _, form_body = req_body:match(
+    local payload_headers, form_body = req_body:match(
         boundary_line .. '\r\n' ..
         '(.-\r\n)' .. '\r\n' .. -- headers
         '(.-)' .. '\r\n' .. -- body
         boundary_line
     )
-    return form_body
+
+    local filename, _
+    if payload_headers ~= nil then
+        _, filename = payload_headers:match('Content%-Disposition%: form%-data; name="(.-)"; filename="([^"]-)"')
+    end
+
+    return form_body, nil, {
+        filename = filename,
+    }
 end
 
 --- Wait for box.info.vclock[id] to reach desired LSN.
--- a/test/unit/utils_test.lua
+++ b/test/unit/utils_test.lua
@@ -34,3 +34,42 @@ function g.test_randomize_path()
 
     digest.urandom = urandom_original
 end
+
+function g.test_http_read_body()
+    local body = '--c187dde3e9318fcc6509f45f76a89424\r\n'..
+    'Content-Disposition: form-data; name="file"; filename="sample.txt"\r\n' ..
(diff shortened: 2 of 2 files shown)


test/unit/utils_test.lua
```diff
@@ -34,3 +34,24 @@ function g.test_randomize_path()
...
+        read = function()
+            return body
+        end,
+        headers = {
+            ['content-type'] = 'multipart/form-data; boundary=c187dde3e9318fcc6509f45f76a89424',
+        }
+    }
+    local payload, err, meta = utils.http_read_body(req)
+    t.assert_equals(payload, 'Content file')
+    t.assert_equals(err, nil)
```
yngvar-antonsson, line 55:
```suggestion
    t.assert_not(err)
```

test/unit/utils_test.lua
```diff
@@ -34,3 +34,24 @@ function g.test_randomize_path()
...
     digest.urandom = urandom_original
 end
+
+function g.test_http_read_body()
+    local body = '--c187dde3e9318fcc6509f45f76a89424\r\n'..
+'Content-Disposition: form-data; name="file"; filename="sample.txt"\r\n' ..
+'Content-Type: text/plain\r\n' ..
+'\r\n' ..
+'Content file\r\n' ..
+'--c187dde3e9318fcc6509f45f76a89424--'
```
yngvar-antonsson, line 44:
```suggestion
    'Content-Disposition: form-data; name="file"; filename="sample.txt"\r\n' ..
    'Content-Type: text/plain\r\n' ..
    '\r\n' ..
    'Content file\r\n' ..
    '--c187dde3e9318fcc6509f45f76a89424--'
```

cartridge/utils.lua
```diff
@@ -356,13 +356,21 @@ local function http_read_body(req)
     -- that have a special meaning with % to escape them.
```
yngvar-antonsson, line 356: Could you please add a function description in ldoc style if you have time?

cartridge/utils.lua
```diff
@@ -356,13 +356,21 @@ local function http_read_body(req)
...
     )
-    return form_body
+
+    local filename, _
+    if payload_headers ~= nil then
+        _, filename = payload_headers:match('Content%-Disposition%: form%-data; name="(.-)"; filename="([^"]-)"')
+    end
+
+    return form_body, nil, {
+        filename = filename,
```
yngvar-antonsson, line 372: Why do you return filename inside table?

cartridge/utils.lua
```diff
@@ -356,13 +356,21 @@ local function http_read_body(req)
...
-    local _, form_body = req_body:match(
+    local payload_headers, form_body = req_body:match(
         boundary_line .. '\r\n' ..
         '(.-\r\n)' .. '\r\n' .. -- headers
         '(.-)' .. '\r\n' .. -- body
         boundary_line
     )
-    return form_body
+
+    local filename, _
```
yngvar-antonsson, line 366:
```suggestion
    local filename
```

yngvar-antonsson (changes requested):

cartridge/utils.lua
olegrok: No. Otherwise _ will be considered as undefined variable

cartridge/utils.lua
olegrok: It's an approach that allows to extend API. Probably some other headers will be also required and it will be possible to add just new field to a table.

test/unit/utils_test.lua
```diff
@@ -34,3 +34,24 @@ function g.test_randomize_path()
 
     digest.urandom = urandom_original
 end
+
+function g.test_http_read_body()
+    local body = '--c187dde3e9318fcc6509f45f76a89424\r\n'..
+'Content-Disposition: form-data; name="file"; filename="sample.txt"\r\n' ..
```
Steap2448, line 40: If I remember correctly, filename is an optional directive. Could you cover a case when there is none?

Steap2448 (changes requested):

cartridge/utils.lua
olegrok: Added

test/unit/utils_test.lua
olegrok: Applied

test/unit/utils_test.lua
olegrok: Applied

test/unit/utils_test.lua
olegrok: Good point. Added.

yngvar-antonsson (approved):

PR #1616, opened 2021-11-25
Add `Cluster:server_by_role` method

### LINKED ISSUE #1615: Getting a connection to the server only by the name of the role in the test cluster
In `cartrdige.rpc` there is a `get_connection` method that only gets a connection to the server by role. I want a similar method for working with a test cluster.

palage4a:
## What has been done? Why? What problem is being solved?
Added a method that allows you to get a server from a test cluster by the name of the role that is applied on the server.

I didn't forget about

- [X] Tests
- [ ] Changelog
- [X] Documentation

Close #1615

### DIFF
--- a/cartridge/test-helpers/cluster.lua
+++ b/cartridge/test-helpers/cluster.lua
@@ -9,6 +9,7 @@ local log = require('log')
 local uuid = require('uuid')
 
 local luatest = require('luatest')
+local utils = require('cartridge.utils')
 local Server = require('cartridge.test-helpers.server')
 local Stateboard = require('cartridge.test-helpers.stateboard')
 
@@ -159,15 +160,9 @@ end
 --- Return iterator for cluster server's with enabled role
 local function iter_servers_by_role(cluster, role_name)
     local replicasets_with_role = fun.iter(cluster.replicasets)
-    :filter(function(rs)
-        return fun.iter(rs.roles)
-            :filter(function(role)
-                return role == role_name
-            end)
-            :nth(1)
-    end)
-    :map(function(rs) return rs.uuid, true end)
-    :tomap()
+        :map(function(rs) return rs.uuid, rs.roles end)
+        :map(function(rs_uuid, roles) return rs_uuid, utils.table_find(roles, role_name) ~= nil end)
+        :tomap()
 
     return fun.iter(cluster.servers)
         :filter(function(server)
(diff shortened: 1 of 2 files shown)


yngvar-antonsson (approved):

cartridge/test-helpers/cluster.lua
```diff
@@ -156,6 +156,39 @@ function Cluster:server(alias)
...
+            :nth(1)
+    end)
+    :map(function(rs) return rs.uuid, true end)
+    :tomap()
+
+    return fun.iter(cluster.servers)
+        :filter(function(server)
+            return replicasets_with_role[server.replicaset_uuid]
+        end)
+end
```
olegrok, line 176: To be honest patch looks like you first time faced luafun and tried to use it everywhere.

It could be easily rewritten as
```lua
local function iter_servers_by_role(cluster, role_name)
    local replicasets_with_role = fun.iter(cluster.replicasets)
        :map(function(rs) return rs.uuid, rs.roles end)
        :map(function(rs_uuid, roles) return rs_uuid, utils.table_find(roles, role_name) ~= nil end)
        :tomap()

    return fun.iter(cluster.servers)
        :filter(function(server)
            return replicasets_with_role[server.replicaset_uuid]
        end)
end
```

For me it's more clear and laconic.

olegrok: Also I hope you won't write production code in such way. Huge amount of closures is a perfect performance killer.

palage4a: > you first time faced luafun and tried to use it everywhere

`luafun` is not the only tool that allows you to use FP :)
But you'r right, its first time.

> Also I hope you won't write production code in such way. Huge amount of closures is a perfect performance killer.

Thanks for your advice. I like your version a little more, but this does not affect the execution speed in any way.
For test it i increased count of replicasets in helper cluster by 1k.
PS
> Huge amount of closures is a perfect performance killer

I agree with that,  but you have the same number of closures as me, dont you?

olegrok: > I agree with that, but you have the same number of closures as me, dont you?

It's not so critical for tests. I told about production code.

yngvar-antonsson: @palage4a would you apply that refactoring?

palage4a: > @palage4a would you apply that refactoring?

yes, sorry for delay. I will try update PR today.

olegrok (approved):
LGTM

yngvar-antonsson (approved):

Issue #1372, opened 2021-04-08
Call box.ctl.clear_synchro_queue() on new leader


R-omk:
The external algorithm of  leader selection  needs to take into account the specifics of synchronous replication.

https://github.com/tarantool/doc/issues/1964
https://github.com/tarantool/tarantool/issues/5435

R-omk:
related https://github.com/tarantool/cartridge/issues/1233

yngvar-antonsson:
For the future development:
- sync replication should work with any failover type
- so we need to set election mode to manual it case of eventual/stateful failover and call box.ctl.promote on leader change

filonenko-mikhail:
The eventual failover is not compatible with box.ctl.promote()

PR #1622, opened 2021-12-02
Raft in Cartridge

### LINKED ISSUE #1233: ☂️ Support native tarantool election_mode as cartridge failover mode
_Modified by @filonenko-mikhail_

## Motivation
There are two failover modes in catridge:
  - eventual
  - stateful
  
Eventual is lightweight but not strong consistent.
Stateful consistent but requires additional sync cross dc kv storage cluster.

Introduce new mode for built-in raft based tarantool failover.

## Bucket Ref On Replica Problem 

Case, when bucket ref was on replica and read the query from replica, lost data during rebalance happened.

- https://github.com/tarantool/vshard/issues/173

## Server disable/enable

The problem when raft switch ro, at this time vshard does not know that request routing changed. vshard router knows about new raft master only after failover timeout + reconfiguration timeout.

If `storage.disable` exists, it happens right after raft switch master flag.

- https://github.com/tarantool/vshard/discussions/307

## We do not know by raft monitoring about leader lost

For e.g. there are three nodes. The first is a leader. And there are two followers.
First two become dead. And the third remember that leader is first. And we can't make the right decision according to the third node world view.

- https://github.com/tarantool/tarantool/issues/6690

## Auto revoke leadership when connectivity loss (r-omk)

- https://github.com/tarantool/tarantool/issues/6661

## (probably) Split-brain when low quorum

- https://github.com/tarantool/tarantool/issues/5295

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1233 

Known issues:
- Will not work in case when there are less than 3 nodes in replicaset.
- Could be useful: https://github.com/tarantool/tarantool/commit/e1d2f7f0073cdd893e3bdd032bd48bad33e156bf

TODO:
- Call box.ctl.promote in eventual and stateful mode

### DIFF
--- a/cartridge/failover.lua
+++ b/cartridge/failover.lua
@@ -40,6 +40,7 @@ local topology = require('cartridge.topology')
 local service_registry = require('cartridge.service-registry')
 local stateboard_client = require('cartridge.stateboard-client')
 local etcd2_client = require('cartridge.etcd2-client')
+local raft_failover = require('cartridge.failover.raft')
 
 local FailoverError = errors.new_class('FailoverError')
 local SwitchoverError = errors.new_class('SwitchoverError')
@@ -47,6 +48,7 @@ local ApplyConfigError = errors.new_class('ApplyConfigError')
 local ValidateConfigError = errors.new_class('ValidateConfigError')
 local StateProviderError = errors.new_class('StateProviderError')
 
+vars:new('mode')
 vars:new('all_roles')
 vars:new('instance_uuid')
 vars:new('replicaset_uuid')
@@ -661,6 +663,11 @@ local function cfg(clusterwide_config)
     local failover_cfg = topology.get_failover_params(topology_cfg)
     local first_appointments
 
+    -- disable raft if it was enabled
+    if vars.mode == 'raft' and failover_cfg.mode ~= 'raft' then
+        raft_failover.disable()
+    end
+
     if failover_cfg.mode == 'disabled' then
         log.info('Failover disabled')
         vars.fencing_enabled = false
@@ -753,6 +760,25 @@ local function cfg(clusterwide_config)
             end,
         })
         vars.failover_fiber:name('cartridge.stateful-failover')
+    elseif failover_cfg.mode == 'raft' then
+        local ok, err = ApplyConfigError:pcall(raft_failover.cfg)
+        if not ok then
+            return nil, err
+        end
+
+        vars.fencing_enabled = false
+        vars.consistency_needed = false
+        first_appointments = raft_failover.get_appointments(topology_cfg)
+
+        vars.failover_fiber = fiber.new(failover_loop, {
+            get_appointments = function()
+                vars.membership_notification:wait()
+                return raft_failover.get_appointments(topology_cfg)
+            end,
+        })
+        vars.failover_fiber:name('cartridge.raft-failover')
+
+        log.info('Raft failover enabled')
     else
         return nil, ApplyConfigError:new(
             'Unknown failover mode %q',
@@ -789,6 +815,8 @@ local function cfg(clusterwide_config)
     box.cfg({
         read_only = not vars.cache.is_rw,
     })
+
+    vars.mode = failover_cfg.mode
 
     return true
 end
--- a/cartridge/failover/raft.lua
+++ b/cartridge/failover/raft.lua
@@ -1,0 +1,125 @@
+local membership = require('membership')
+local checks = require('checks')
+local log = require('log')
+local fun = require('fun')
+
+local topology = require('cartridge.topology')
+local pool = require('cartridge.pool')
+local argparse = require('cartridge.argparse')
+
+local vars = require('cartridge.vars').new('cartridge.failover')
+local errors = require('errors')
+
+local PromoteLeaderError = errors.new_class('PromoteLeaderError')
+local AppointmentError = errors.new_class('AppointmentError')
+
+vars:new('leader_uuid')
+vars:new('raft_trigger')
+
+--- Generate appointments according to raft status.
+-- Used in 'raft' failover mode.
+-- @function get_appointments
+local function get_appointments(topology_cfg)
+    checks('table')
+    local replicasets = assert(topology_cfg.replicasets)
+
+    local appointments = {}
+
+    local box_info = box.info()
+    for replicaset_uuid, _ in pairs(replicasets) do
+        local leaders = topology.get_leaders_order(
+            topology_cfg, replicaset_uuid
+        )
+        if replicaset_uuid == vars.replicaset_uuid then
+            local my_leader_id = box_info.election.leader
+            local my_leader = box_info.replication[my_leader_id]
+            if my_leader ~= nil then
+                appointments[replicaset_uuid] = my_leader.uuid
+                goto next_rs
+            end
+        end
+
+        local latest_leader
+        local latest_term = 0
+        for _, instance_uuid in ipairs(leaders) do
+            local server = topology_cfg.servers[instance_uuid]
+            local member = membership.get_member(server.uri)
+
+            if member ~= nil
+            and member.payload.leader_uuid ~= nil
+            and member.payload.raft_term or 0 >= latest_term
+            then
+                latest_leader = member.payload.leader_uuid
+                latest_term = member.payload.raft_term or 0
+            end
+        end
+        appointments[replicaset_uuid] = latest_leader
+        ::next_rs::
+    end
+
+    appointments[vars.replicaset_uuid] = vars.leader_uuid
+    return appointments
+end
+
+local function on_election_trigger()
+    local box_info = box.info()
+    local election = box_info.election
+
+    local leader = box_info.replication[election.leader] or {}
+
+    if vars.leader_uuid ~= leader.uuid then
+        vars.cache.is_leader = vars.leader_uuid == vars.instance_uuid
+        vars.leader_uuid = leader.uuid
+        membership.set_payload('leader_uuid', vars.leader_uuid)
+    end
+    membership.set_payload('raft_term', election.term)
+end
+
+local function cfg()
+    assert(box.ctl.on_election, "Your Tarantool version doesn't support raft failover mode")
+    log.warn('Raft failover is in beta-version')
+    local box_opts = argparse.get_box_opts()
+
+    box.cfg{
+        -- The instance is set to candidate, so it may become the leader
+        -- as well as vote for other instances.
+        --
+        -- Alternative: set one of instances to `voter` so that it
+        -- never becomes a leader but still votes for one of its peers and helps
+        -- it reach election quorum.
+        election_mode = box_opts.election_mode or 'candidate',
+        -- Quorum for both synchronous transactions and
+        -- leader election votes.
+        replication_synchro_quorum = box_opts.replication_synchro_quorum or 'N/2 + 1',
+        -- Synchronous replication timeout. The transaction will be
+        -- rolled back if no quorum is achieved during timeout.
+        replication_synchro_timeout = box_opts.replication_synchro_timeout,
+        -- Timeout between elections. Needed to restart elections when no leader
+        -- emerges soon enough. Equals 4 * replication_timeout
+        election_timeout = box_opts.election_timeout,
+    }
+
+    if vars.raft_trigger == nil then
+        vars.raft_trigger = box.ctl.on_election(on_election_trigger)
+    end
+
+    membership.set_payload('raft_term', box.info.election.term)
+
+    return true
+end
+
+-- disable raft if it was enabled
+local function disable()
+    if vars.raft_trigger ~= nil then
+        box.ctl.on_election(nil, vars.raft_trigger)
+        vars.raft_trigger = nil
+    end
+    box.cfg{ election_mode = 'off' }
+    vars.leader_uuid = nil
+end
+
+return {
+    cfg = cfg,
+    disable = disable,
+    get_appointments = get_appointments,
+}
--- a/cartridge/failover/raft.lua
+++ b/cartridge/failover/raft.lua
@@ -97,6 +97,11 @@ local function cfg()
         -- Timeout between elections. Needed to restart elections when no leader
         -- emerges soon enough. Equals 4 * replication_timeout
         election_timeout = box_opts.election_timeout,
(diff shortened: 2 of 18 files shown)


cartridge/failover.lua
```diff
@@ -730,6 +804,57 @@ local function cfg(clusterwide_config)
...
+            -- The instance is set to candidate, so it may become leader itself
+            -- as well as vote for other instances.
+            --
+            -- Alternative: set one of the three instances to `voter`, so that it
+            -- never becomes a leader but still votes for one of its peers and helps
+            -- it reach election quorum (2 in our case).
+            election_mode = os.getenv('TARANTOOL_ELECTION_MODE') or 'candidate',
+            -- Quorum for both synchronous transactions and
+            -- leader election votes.
+            replication_synchro_quorum = failover_cfg.replication_synchro_quorum or 2,
```
R-omk, line 820: Since 2.6.2  the option can be expression string
`replication_synchro_quorum = "N/2 + 1"`

cartridge/failover.lua
```diff
@@ -730,6 +804,57 @@ local function cfg(clusterwide_config)
...
+            -- Timeout between elections. Needed to restart elections when no leader
+            -- emerges soon enough.
+            election_timeout = 0.25, -- = 4 * repl_timeout
+        }
+
+        local election = box.info.election
+
+        vars.raft_term = election.term
+
+        local leader = box.info.replication[election.leader] or {}
```
sergepetrenko, line 837: Leader won't have time to elect here, so it'll be `{}`.

Is it possible to yield here until leader is found?

test/integration/raft_test.lua
```diff
@@ -0,0 +1,323 @@
...
+    end):tomap()
+end
+
+g.before_each(function()
+    h.retrying({}, function()
+        t.assert_equals(h.list_cluster_issues(g.cluster.main_server), {})
+    end)
+    h.retrying({}, function()
+        -- call box.ctl.promote on storage-1
+        set_master('storage-1')
```
sergepetrenko, line 158: It's not guaranteed that 'storage-1' will be master after calling promote on it.
`box.ctl.promote` runs elections, and it is possible (though highly unlikely), that some other node wins them.
In case the node hasn't won the elections, promote will return an error.

So you may want to do something like `while not ok do ok = pcall(box.ctl.promote) end` to get the expected behaviour.

Or, as an alternative, set everyone except the promoted node to be `follower`.

sergepetrenko (commented):
Thanks for the patch!
Please see my comments below.

filonenko-mikhail:
# Related issues:

## Bucket Ref On Replica Problem 

Case when bucket ref was on replica and read query from replica lost data during rebalance happened

- https://github.com/tarantool/vshard/issues/173

cartridge/failover.lua
```diff
@@ -730,6 +735,22 @@ local function cfg(clusterwide_config)
...
+    elseif failover_cfg.mode == 'raft' then
+        raft_failover.cfg(failover_cfg)
+
+        vars.fencing_enabled = false
+        vars.consistency_needed = false
+        first_appointments = raft_failover.get_appointments(topology_cfg)
+
+        vars.failover_fiber = fiber.new(failover_loop, {
+            get_appointments = function()
+                vars.membership_notification:wait()
```
olegrok, line 747: Do I correctly understand that any notification will be considered as leader change? Why is it not enough to transfer this info via channel from election trigger? Instances in other replicasets have its own leaders that shouldn't be affected if leader of single replicaset change.

cartridge/failover/raft.lua
```diff
@@ -0,0 +1,118 @@
...
+    box.cfg{ election_mode = 'off' }
+    vars.raft_term = nil
+    vars.leader_uuid = nil
+end
+
+return {
+    cfg = cfg,
+    disable = disable,
+    get_appointments = get_appointments,
+}
```
olegrok, line 118: Manual promotion should be also available. See box.ctl.promote.

cartridge/failover/raft.lua
```diff
@@ -0,0 +1,118 @@
...
+    assert(box.ctl.on_election, "Raft failover is available only in Tarantool >= 2.10.0")
+
+    box.cfg{
+        -- The instance is set to candidate, so it may become leader itself
+        -- as well as vote for other instances.
+        --
+        -- Alternative: set one of instances to `voter`, so that it
+        -- never becomes a leader but still votes for one of its peers and helps
+        -- it reach election quorum.
+        election_mode = os.getenv('TARANTOOL_ELECTION_MODE') or 'candidate',
```
olegrok, line 80: I guess argparse should be used instead of "os.getenv"

test/integration/raft_test.lua
```diff
@@ -0,0 +1,474 @@
...
+end
+
+g.after_all = function()
+    g.cluster:stop()
+    fio.rmtree(g.cluster.datadir)
+end
+
+local function set_master(instance_name)
+    g.cluster:server(instance_name):exec(function()
+        box.ctl.promote()
```
olegrok, line 127: I don't see any "box.ctl.promote" call in failover code. Should user call it manually?

test/integration/raft_test.lua
```diff
@@ -0,0 +1,474 @@
...
+        uuid = single_replicaset_uuid,
+    })
+
+    -- send bucket to another storage
+    local bucket_counts = g.cluster:server('storage-1'):exec(function(bucket_id, replicaset_uuid)
+        assert(box.info.ro ~= true)
+        local vshard_storage = require('vshard.storage')
+        vshard_storage.bucket_send(bucket_id, replicaset_uuid)
+        -- wait until sending
+        require('fiber').sleep(0.5)
```
olegrok, line 449: Why not "wait_for"? Sleeps in code are quite unreliable. Especially on CI where a lot of tasks/containers are run simultaneously.

test/integration/raft_test.lua
```diff
@@ -0,0 +1,474 @@
+local fio = require('fio')
```
ligurio, line 1: This is a good test suite to start, however, these tests are not enough for testing failover. I would recommend adding  randomness to tests. For example, create a test where we write and read to cluster and periodically kill a leader. Data on cluster must be in a consistent state. Such test for Tarantool was very effective (see https://github.com/tarantool/tarantool/issues/5055#issuecomment-778166416).

cartridge/failover.lua
```diff
@@ -730,6 +735,22 @@ local function cfg(clusterwide_config)
             end,
         })
         vars.failover_fiber:name('cartridge.stateful-failover')
+    elseif failover_cfg.mode == 'raft' then
+        raft_failover.cfg(failover_cfg)
+
+        vars.fencing_enabled = false
+        vars.consistency_needed = false
```
ligurio, line 742: Does it mean that we don't need consistency when Raft is enabled?

ligurio (changes requested):

doc/schema.graphql
```diff
@@ -384,6 +256,7 @@ type MutationApicluster {
     """The `Boolean` scalar type represents `true` or `false`."""
     enabled: Boolean
 
+<<<<<<< HEAD
```
filonenko-mikhail, line 259: dirty merge

cartridge/failover.lua
yngvar-antonsson: The current failover mechanism needs to know who is a leader in other replicasets to use this info for rpc calls. Now eventual (and Raft, based on eventual) receives notifications about leader changes and use it to update local leader list.

yngvar-antonsson: The current failover mechanism needs to know who is a leader in other replicasets to use this info for rpc calls. Now eventual (and Raft, based on eventual) receives notifications about leader changes and use it to update local leader list.

olegrok: Any leader change anyway triggers apply_config. I guess topology should be enough.

cartridge/failover.lua
```diff
@@ -753,6 +760,25 @@ local function cfg(clusterwide_config)
             end,
         })
         vars.failover_fiber:name('cartridge.stateful-failover')
+    elseif failover_cfg.mode == 'raft' then
+        local ok, err = ApplyConfigError:pcall(raft_failover.cfg)
+        if not ok then
+            return nil, err
+        end
+
+        vars.fencing_enabled = false
```
olegrok, line 769: After https://github.com/tarantool/tarantool/commit/ed46feb27087247ab5ee47fcf63f28c3ad842099 there is `election_fencing_enabled` option. I guess it should be supported.

cartridge/failover/raft.lua
```diff
@@ -0,0 +1,137 @@
...
+    local replicasets = assert(topology_cfg.replicasets)
+
+    local appointments = {}
+
+    for replicaset_uuid, _ in pairs(replicasets) do
+        local leaders = topology.get_leaders_order(
+            topology_cfg, replicaset_uuid
+        )
+        if replicaset_uuid == vars.replicaset_uuid then
+            local my_leader_id = box.info.election.leader
```
olegrok, line 28: nit: box.info is not so cheap call. Probably it's better to call it once. (it's not serious because it's not hot path).

cartridge/failover/raft.lua
```diff
@@ -0,0 +1,137 @@
...
+            local my_leader = box.info.replication[my_leader_id]
+            if my_leader ~= nil then
+                appointments[replicaset_uuid] = my_leader.uuid
+                goto next_rs
+            end
+        end
+
+        local last_leader
```
olegrok, line 36: last or latest?

cartridge/failover/raft.lua
```diff
@@ -0,0 +1,137 @@
...
+        for _, instance_uuid in ipairs(leaders) do
+            local server = topology_cfg.servers[instance_uuid]
+            local member = membership.get_member(server.uri)
+
+            if member ~= nil
+            and member.payload.leader_uuid ~= nil
+            and member.payload.raft_term >= last_term
+            then
+                last_leader = member.payload.leader_uuid
+                last_term = member.payload.raft_term
```
olegrok, line 47: Is it possible that current instance will have "last_term"? Should it be handled anyhow?

olegrok: And if not - why not? I guess comment required.

cartridge/failover/raft.lua
```diff
@@ -0,0 +1,137 @@
...
+        end
+        appointments[replicaset_uuid] = last_leader
+        ::next_rs::
+    end
+
+    appointments[vars.replicaset_uuid] = vars.leader_uuid
+    return appointments
+end
+
+local function on_election_trigger()
```
olegrok, line 58: I guess this function should trigger failover. Still I'm not sure that failover based on raft should use "eventual" mechanisms. It kills all benefits. Are there any spec/RFC? Who and why suggested it and decided that it should be work so?

cartridge/failover/raft.lua
```diff
@@ -0,0 +1,137 @@
...
+    local election = box.info.election
```
olegrok, line 59: nit: the same comment about box.info.

cartridge/failover/raft.lua
```diff
@@ -0,0 +1,137 @@
...
+        -- Quorum for both synchronous transactions and
+        -- leader election votes.
+        replication_synchro_quorum = box_opts.replication_synchro_quorum or 'N/2 + 1',
+        -- Synchronous replication timeout. The transaction will be
+        -- rolled back if no quorum is achieved during timeout.
+        replication_synchro_timeout = box_opts.replication_synchro_timeout,
+        -- Heartbeat timeout. A leader is considered dead if it doesn't
+        -- send heartbeats for 4 * replication_timeout.
+        -- Once the leader is dead, remaining instances start a new election round.
+        replication_timeout = box_opts.replication_timeout,
```
olegrok, line 93: This options are not related to raft failover. It's relevant for syncro replication. But it could work with any type failover. That failover just should call "box.ctl.promote" on leader instance.

cartridge/failover/raft.lua
```diff
@@ -0,0 +1,137 @@
...
+    if vars.raft_trigger ~= nil then
+        box.ctl.on_election(nil, vars.raft_trigger)
+        vars.raft_trigger = nil
+    end
+    box.cfg{ election_mode = 'off' }
+    vars.leader_uuid = nil
+end
+
+local function promote(replicaset_leaders)
+    local servers_list = package.loaded.cartridge.confapplier.get_readonly('topology').servers
```
olegrok, line 119: I'm not sure but in case of advertise_uri change real value could be different. And "refine_uri..." function should be used.

cartridge/failover.lua
R-omk: Suppose there are many instances running on the same host, what happens if the host goes missing?  When for each leader that was on this host, a new one will be elected at the same time, will there not be a race with the loss of actual leader on the other replicasets, or even worse, a  deadloop with a two-phase cfg apply commit?

Everything looks like it need to somehow prevent the massive simultaneous applying the config.

olegrok (changes requested):
This patch doesn't "Close https://github.com/tarantool/cartridge/issues/1372". There is only one mention of box.ctl.promote in your patch.

Remove this "close" from commit messages/PR description or finally implement it. It ~~should~~ must be possible to use synchro with any type failover.

cartridge/failover.lua
olegrok: > Everything looks like it need to somehow prevent the massive simultaneous applying the config.

It's relevant for any failover type... And probably it's even implemented.

olegrok:
Following testcase should be covered:
  - Enable raft failover mode
  - Disable raft failover mode (or change it, choose e.g. eventual failover mode) - all raft related box-options should be switched off (in fact mode should be changed to "manual"). E.g. we shouldn't have two failovers working at the same time.

cartridge/failover.lua
yngvar-antonsson: As we discussed IRL, leader table not in the topology, so we need membership to notify instances about leader change

filonenko-mikhail (approved):

olegrok (changes requested):
24 commits can't be pushed as is. Please split patch to atomic changes and add detailed description for each commit.

Patch is too huge and important and should have good history/documentation.

olegrok (approved):
I left several minor comments below.

PR #1690, opened 2021-12-30
Fix GraphQL query `auth_params.username`

### LINKED ISSUE #1633: [2pt] Make field "Full name" obligatory on the form "Add a new user"
To prevent situation when authorized user is working with empty value in the right top corner

image

usenko-timur:
* Fix GraphQL query ``auth_params.username`` returns empty string instead of ``username``.

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1633

### DIFF
--- a/cartridge/webui/api-auth.lua
+++ b/cartridge/webui/api-auth.lua
@@ -76,7 +76,7 @@ local function get_auth_params()
 
     local username = auth.get_session_username()
     local user = username and auth.get_user(username)
-    if user ~= nil and user.fullname ~= nil then
+    if user ~= nil and user.fullname ~= nil and #user.fullname > 0 then
         username = user.fullname
     end
     return {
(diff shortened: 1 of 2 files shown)


yngvar-antonsson (approved):

cartridge/webui/api-auth.lua
```diff
@@ -76,7 +76,7 @@ local function get_auth_params()
 
     local username = auth.get_session_username()
     local user = username and auth.get_user(username)
-    if user ~= nil and user.fullname ~= nil then
+    if user and (user.fullname or '') ~= '' then
```
olegrok, line 79: This check is quite strange.
  - Check will be broken if fullname is box.NULL
  - And seems it's easier to check that string length has non-zero length: `user ~= nil and user.fullname ~= nil and #user.fullname > 0`

usenko-timur: Thanks, fixed!

olegrok (approved):

OlgaYakovleva (approved):
Verified on cartridge as on TDG2 there is not filed "full name"

PR #1692, opened 2022-01-10
disable vshard.storage in case of operation error

### LINKED ISSUE #1411: Storages with OperationError state should be excluded from vshard config
Since vshard doesn't know anything about cartridge state machine we could have a case when
vshard makes requests to instance with OperationError state. It looks wrong and unclear for customers.

olegrok:
In case of OperationError (config was unsuccessfully applied on
storage) we shouldn't perform request to such storage.
After this feature was implemented in vshard
(https://github.com/tarantool/vshard/issues/298) we could just
disable vshard storage on such instances. For this purpose simple
trigger on_apply_config was implemented.

Closes #1411


olegrok:
Currently it could be considered as PoC. Blocked until https://github.com/tarantool/vshard/issues/310.

yngvar-antonsson (approved):

olegrok:
However, patch is a bit incomplete. Apply_config could be run by failover. And this case should be added and tested

cartridge/failover.lua
```diff
@@ -305,6 +305,20 @@ local function apply_config(mod)
...
+local function on_apply_config(mod, state)
+    checks('?table', 'string')
+    if mod == nil then
+        return true
+    end
+
+    local conf = vars.clusterwide_config:get_readonly()
+
+    if type(mod.on_apply_config) == 'function' then
+        ApplyConfigError:pcall(mod.on_apply_config, conf, state)
```
olegrok, line 318: Probably it makes sense to log here in case of error

CHANGELOG.rst
```diff
@@ -31,6 +31,8 @@ Added
 
 - Tarantool Raft options are supported in argparse.
 
+- ``on_apply_config(conf, state)`` roles API.
```
olegrok, line 34: Let's stay it for internal usage. Nobody asked about such features. Let's stay it in "gray" zone and won't extend existing API

olegrok:
LGTM.
Feel free to merge or someone for the second review round.

PR #1717, opened 2022-01-22
Restrict RPC calls to work while clusterwide config

### LINKED ISSUE #1575: Fix `rpc.get_candidates` function

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog

Close #1575

### DIFF
--- a/cartridge/rpc.lua
+++ b/cartridge/rpc.lua
@@ -98,7 +98,7 @@ local function get_candidates(role_name, opts)
     end
 
     local end_time = fiber.time() + opts.retry_timeout
-    local patch_in_progress = true
+    local patch_in_progress
     repeat
         patch_in_progress = assert(twophase_vars.locks)['clusterwide']
         if not patch_in_progress then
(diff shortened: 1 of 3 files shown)


cartridge/rpc.lua
```diff
@@ -76,17 +78,40 @@ local function get_candidates(role_name, opts)
...
+        patch_in_progress = (twophase_vars.locks or {})['clusterwide']
+        if not patch_in_progress then
+            break
+        else
+            fiber.sleep(opts.retry_delay)
+        end
+    until fiber.time() <= end_time
+
+    if patch_in_progress then
+        return nil, RemoteCallError:new('Twophase commit is in progress, try again later')
```
olegrok, line 112: All TDG tests fail after this patch. Please don't do that.

olegrok (changes requested):

cartridge/rpc.lua
olegrok: I pushed my "optimistic" approach https://github.com/tarantool/cartridge/pull/1724

I have not tested it against TDG but seems it doesn't ban anything just add some timeout to wait when config commit will be finished.

yngvar-antonsson:
Closed in favor of #1724

PR #1724, opened 2022-01-23
rpc: wait until config commit finishes if role is unavailable

### LINKED ISSUE #1575: Fix `rpc.get_candidates` function

olegrok:
There is a raise condition. We could commit config locally but it
could be in progress on some instance. Before this patch user got
unexpected "Role X unavailable" from instance where such role was
assumed.
Solution is an optimistic approach - detect config apply and try
to wait until it will be finished.

Closes #1575

### DIFF
--- a/cartridge/rpc.lua
+++ b/cartridge/rpc.lua
@@ -14,6 +14,7 @@ local roles = require('cartridge.roles')
 local topology = require('cartridge.topology')
 local failover = require('cartridge.failover')
 local confapplier = require('cartridge.confapplier')
+local twophase = require('cartridge.twophase')
 local service_registry = require('cartridge.service-registry')
 
 local RemoteCallError = errors.new_class('RemoteCallError')
@@ -21,6 +22,16 @@ local RemoteCallError = errors.new_class('RemoteCallError')
 local function call_local(role_name, fn_name, args)
     checks('string', 'string', '?table')
     local role = service_registry.get(role_name)
+
+    if role == nil then
+        -- Optimistic approach.
+        -- Probably instance in apply config state so
+        -- we need to wait a bit until it finishes.
+        if twophase.wait_config_release()
+        and confapplier.wish_state('RolesConfigured') == 'RolesConfigured' then
+            role = service_registry.get(role_name)
+        end
+    end
 
     if role == nil then
         return nil, RemoteCallError:new('Role %q unavailable', role_name)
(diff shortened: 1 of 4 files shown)


yngvar-antonsson:
Have you tested it on TDG?

olegrok:
> Have you tested it on TDG?

It seems to be ok.
However I found one more issue with #1575 that this patch is not fix.

If there are three instances A, B, C and A applies config and disables some role on C. If B still has old version of configuration (not committed yet) and thinks that C has role that is actually disabled we will still have such error "Role unavailable".

So patch fixes only one (probably the most often) flow.

test/integration/get_candidates_test.lua
```diff
@@ -0,0 +1,95 @@
...
+           res = res,
+           err = err,
+           candidates = candidates,
+       }
+   end)
+
+    t.assert_type(res.err, 'nil')
+    t.assert_equals(res.res, 'woof')
+    t.assert_equals(res.candidates, {g.B1.advertise_uri})
+end
```
yngvar-antonsson, line 95: Add newline

cartridge/rpc.lua
```diff
@@ -22,6 +23,16 @@ local function call_local(role_name, fn_name, args)
     checks('string', 'string', '?table')
     local role = service_registry.get(role_name)
 
+    if role == nil then
+        -- Optimistic approach.
+        -- Probably instance in apply config state so
+        -- we need to wait a bit until it finishes.
+        if twophase.wait_config_release() and
+                confapplier.wish_state('RolesConfigured') == 'RolesConfigured' then
```
yngvar-antonsson, line 31:
```suggestion
        if twophase.wait_config_release()
        and confapplier.wish_state('RolesConfigured') == 'RolesConfigured' then
```

cartridge/twophase.lua
```diff
@@ -80,6 +89,26 @@ local function get_apply_config_timeout()
     return vars.options.apply_config_timeout
 end
 
+local function wait_config_release(timeout)
```
yngvar-antonsson, line 92: Maybe we should log a warning while waiting?

yngvar-antonsson, line 92: Please add docs for this function

yngvar-antonsson (changes requested):

cartridge/twophase.lua
olegrok: It's a critical section. We can't just log here because it could spam logs.

olegrok: Done

yngvar-antonsson (approved):

PR #1730, opened 2022-01-24
Add test to rpc_calls from expelled instance

### LINKED ISSUE #1726: Expelled instance has access to cartridge.rpc
I noticed that if a instance is expelled out of the cluster, it will still be able to make `cartridge.rpc` calls, and maybe something else.

This is very bad.

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog

Close #1726

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,7 +33,7 @@ Added
 - Setters ans getters for timeout options in ``twophase.lua`` (#1440):
   ``netbox_call_timeout``, ``upload_config_timeout``, ``validate_config_timeout``, ``apply_config_timeout``.
 
-- New tests cases (#892, #944, #1473).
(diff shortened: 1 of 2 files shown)


cartridge/rpc.lua
```diff
@@ -279,7 +286,7 @@ local function call_remote(role_name, fn_name, args, opts)
         return errors.netbox_call(
             conn,
             '_G.__cluster_rpc_call_local',
-            {role_name, fn_name, args},
+            {role_name, fn_name, args, confapplier.get_instance_uuid()},
```
olegrok, line 289: Such approach adds new 36 bytes to each rpc call and makes it slower in general. Why can't we perform such check locally? Doesn't expelled instance know that it's actually expelled?

filonenko-mikhail (approved):

PR #1750, opened 2022-02-11
disable schema fetch for cartridge.pool connections

olegrok:
Cartridge uses only netbox.call and eval. So seems we could safety
disable schema fetching for netbox. New Tarantool versions allows
to disable schema fetching to avoid additional pressure that
schema fetching creates.

See also https://github.com/tarantool/doc/issues/2680


olegrok:
After release 2.10-rc1 itsn't blocked anymore.

filonenko-mikhail (approved):

PR #1760, opened 2022-02-20
avoid recalculations for enable roles list

olegrok:
To check that role is enabled on instance we rebuild a map with
all replicaset roles. It creates additional pressure on GC and
also slowdown rpc.get_candidates function (that is used by
rpc.call and rpc.get_connection functions). To avoid this issue
let's cache a value calculated once and save it into weak table. A
key is a table from topology - that meen it was the same for all
rpc.get_candidates until apply config happens (old config will be
GCed and previous calculations will also disappear from cache.

Simple benchmark on single node shows that such approach is x3
speedups rpc.get_candidates on single node (from ~0.23s to 0.07s for
10k calls).

---
Benchmark itself:
```lua
do
    clock = require('clock')
    rpc = require('cartridge.rpc')
    start = clock.monotonic()

    for i = 1, 1e5 do
        rpc.get_candidates('core')
    end
    print(clock.monotonic() - start)
end
```

### DIFF
--- a/cartridge/roles.lua
+++ b/cartridge/roles.lua
@@ -259,6 +259,9 @@ local function get_enabled_roles_without_deps(roles)
     return ret
 end
 
+-- Local cache to avoid recalculations for the same topology
+local enabled_roles_cache = setmetatable({}, {__mode = 'k'})
+
 --- Roles to be enabled on the server.
 -- This function returns all roles that will be enabled
 -- including their dependencies (both hidden and not)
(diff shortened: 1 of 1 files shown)


yngvar-antonsson:
Please add Changelog

yngvar-antonsson (approved):

olegrok:
> Please add Changelog

There is no user-visible changes. There will be no changelog.

PR #1806, opened 2022-04-25
Allow to suppress failover

### LINKED ISSUE #1758: Stifle failover triggering
# Motivation

A lot of pain across time and humans

## Propolsal

Code name: Steve Stifler

Disable failover when it triggers 10 times in 3 minutes. Enable it again after 5 minutes things fine.

Configuration:
  - enable threshold
  - threshold (10 times in 3 minutes)
  - restore after

UX:
  - user may want reset current stifle any time
  
State of stifler have to be in runtime only, not in clusterwide config, not in filesystem. 
Reset state problably local only (or using rpc call??)

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1758

### DIFF
--- a/cartridge/failover.lua
+++ b/cartridge/failover.lua
@@ -41,6 +41,7 @@ local service_registry = require('cartridge.service-registry')
 local stateboard_client = require('cartridge.stateboard-client')
 local etcd2_client = require('cartridge.etcd2-client')
 local raft_failover = require('cartridge.failover.raft')
+local argparse = require('cartridge.argparse')
 
 local FailoverError = errors.new_class('FailoverError')
 local SwitchoverError = errors.new_class('SwitchoverError')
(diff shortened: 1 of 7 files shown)


cartridge/failover.lua
```diff
@@ -580,6 +589,33 @@ function reconfigure_all(active_leaders)
     confapplier.set_state('RolesConfigured')
 end
 
+--- Lock failover if failover suppressing is on.
+-- @function check_suppressing_lock
+-- @local
+local function check_suppressing_lock()
+    while true do
+        vars.failover_trigger_cnt = 0
+        fiber.sleep(vars.suppress_timeout)
```
filonenko-mikhail, line 598: in case of nil hardcode some const value

filonenko-mikhail (approved):

PR #1812, opened 2022-05-20
Remove non-existing Tarantool 2.9 from tests

yngvar-antonsson:


olegrok:
Still exists https://github.com/tarantool/cartridge/blob/51819872fcd1947cbce4eeea3aeebc0af5255f44/.github/workflows/nightly-test.yml#L19

PR #1822, opened 2022-05-30
Deprecate collect_lua_garbage

### LINKED ISSUE #1814: Deprecate `collect_lua_garbage`
Since `collect_lua_garbage` in vshard 1.20.0 is deprecated, we should deprecate it too

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1814


cartridge/webui/api-vshard.lua
```diff
@@ -32,8 +32,8 @@ local gql_type_vsgroup = gql_types.object({
                 'in the storage group'
         },
         collect_lua_garbage = {
-            kind = gql_types.boolean.nonNull,
-            description = 'If set to true, the Lua collectgarbage() function is called periodically'
+            kind = gql_types.boolean,
+            description = 'Deprecated and has no effect anymore'
         },
```
olegrok, line 37: I suggest to add "deprecationReason" as for collect_bucket_garbage_interval

cartridge/roles/vshard-router.lua
```diff
@@ -161,6 +161,7 @@ local function bootstrap_group(group_name, vsgroup)
 
     local conf = confapplier.get_readonly()
     local vshard_cfg = vshard_utils.get_vshard_config(group_name, conf)
+    vshard_cfg.collect_lua_garbage = nil
```
olegrok, line 164: Do you try to modify part of read-only config here?

cartridge/vshard-utils.lua
```diff
@@ -520,7 +518,7 @@ local function get_vshard_config(group_name, conf)
         rebalancer_max_sending = vshard_groups[group_name].rebalancer_max_sending,
         sched_ref_quota = vshard_groups[group_name].sched_ref_quota,
         sched_move_quota = vshard_groups[group_name].sched_move_quota,
-        collect_lua_garbage = vshard_groups[group_name].collect_lua_garbage,
+        collect_lua_garbage = vshard_groups[group_name].collect_lua_garbage or false,
```
olegrok, line 521: If user specifies `nil` here (or even omits at all) value shouldn't be converted to false

cartridge/roles/vshard-router.lua
yngvar-antonsson: `vshard_utils.get_vshard_config` returns a new table, not read only config

olegrok (approved):

PR #1829, opened 2022-06-02
add test case to check Raft tab

OlgaYakovleva:
add testcase to check RAFT TAB


yngvar-antonsson:
Error has changed: https://github.com/tarantool/cartridge/pull/1830

yngvar-antonsson:
No resources for this patch.

PR #1837, opened 2022-06-08
add utils.is_email_valid tests

olegrok:
This patch adds test for utils.is_email_valid function. Initially
tests were exported from TDG where they passed some fuzzing tests.
Also this way we could keep "utils.is_email_valid" only on
cartridge side and drop that function from TDG.


yngvar-antonsson (approved):

no1seman:
is_email_valid() is not RFC compliant, for example user@localhost is invalid. May be it's time to fix it?

yngvar-antonsson:
@olegrok @no1seman will you make is_email_valid() a RFC compliant or we could merge it as is?

PR #1883, opened 2022-08-21
Add GraphQL examples to docs

yngvar-antonsson:


cartridge/issues.lua
```diff
@@ -55,6 +55,21 @@
 -- * Custom roles can announce more issues with their own level, topic
 --   and message. See `custom-role.get_issues`.
 --
+-- GraphQL request:
+--
+-- You can get info about cluster issues with GrapQL request:
```
patiencedaur, line 60:
```suggestion
-- You can get info about cluster issues using the following GrapQL request:
```

rst/topics/failover.rst
```diff
@@ -301,6 +301,22 @@ Use your favorite GraphQL client (e.g.
 - ``mutation {cluster{failover_params(){}}}``,
 - ``mutation {cluster{failover_promote()}}``.
 
+E.g. how to setup stateful failover:
```
patiencedaur, line 304:
```suggestion
Here is an example of how to setup stateful failover:
```
We [don't use abbreviations like i.e. or e.g.](https://www.tarantool.io/en/doc/latest/contributing/docs/style/#don-t-use-i-e-and-e-g)

PR #1888, opened 2022-08-23
Remote control suspend resume

### LINKED ISSUE #1878: mnt: replication conflict — it is possible that cartridge will be rw on start with opened remote control
There is code to make sure that admin password is the same:

https://github.com/tarantool/cartridge/blob/master/cartridge/confapplier.lua#L580

At the same time remote control is running and can handle remote request also for ddl and dml. 

Probably we have to think how to be in such case.

panticonur:
Introduce `Remote Control` `suspend`/`resume` methods to pause producing requests.

- [x] Tests
- [x] Changelog
- [ ] Documentation

Close #1878

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,8 @@ Added
 
 - Introduced SSL support for Tarantool Enterprise from 2.10.2.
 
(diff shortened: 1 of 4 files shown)


yngvar-antonsson:
Please add test

yngvar-antonsson (approved):

PR #1949, opened 2022-10-12
Save previous member state into state_prev payload and fix member_is_healthy

eugenepaniot:
This change fixes situation when router sends requests to not fully (first start) RolesConfigured node. When node just started it had not configured roles (defined procedures) yet.

* This change saves previous member state into `state_prev` payload;
* The member should be considered as healthy (`member_is_healthy `) if previous state were `RolesConfigured` or `ConfiguringRoles`. According to https://www.tarantool.io/en/doc/latest/book/cartridge/cartridge_dev/#configuringroles transition. Not `BoxConfigured -> ConfiguringRoles`
 

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Close #???

### DIFF
--- a/cartridge/confapplier.lua
+++ b/cartridge/confapplier.lua
@@ -159,6 +159,7 @@ local function set_state(new_state, err)
         )
     end
 
+    membership.set_payload('state_prev', vars.state)
     membership.set_payload('state', new_state)
     vars.state = new_state
     vars.error = err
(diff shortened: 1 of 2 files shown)


yngvar-antonsson:
Please add a Changelog note.

yngvar-antonsson:
And testing would be great.
Thank you for the patch!

yngvar-antonsson (approved):

Issue #1900, opened 2022-09-06
Recalculate leaders after enable raft failover mode


R-omk:
Steps to reproduce problem:

1. Run nodes with raft enabled (`--election-mode candidate`)
2. Launching the cartridge via `cartridge replicasets setup`

The cartridge is now running with disabled failover. 
The first host in the list is selected as non-readonly. 
if we now switch the failover mode to raft, then no matter which node is the leader, the first one from the list will still be selected as non-readonly, in this case, only the force promote can explicitly change the leader.

R-omk:
If enable raft failover before the bootstrap, then the bootstrap will almost certainly fail. And in some replicasets it may pass or not, it leads to partial bootstrap.
It also happens that the "is_master" option pass into apply_config, although in fact the node is in readonly due to another node is the leader.

filonenko-mikhail:
So, it's seems that enabling raft from other than cartridge have to be prohibited.

R-omk:
leader should be computed here (same as raft_term) when raft failover is turning on

https://github.com/tarantool/cartridge/blob/ec98fb9f97f45d2440d84f001cdf7b1fcb6f6b47/cartridge/failover/raft.lua#L128-L130

R-omk:
I believe that all raft data should always be stored to the payload  as one indivisible object  (term number and leader at the same time), otherwise this may lead to the fact that on different nodes the term may change independently of the others.

This can lead to the fact that **different nodes of the same term may contain outdated and different data** about the leader.

https://github.com/tarantool/cartridge/blob/8e40d18d6daa0457268367fe446922ff0d1f09e9/cartridge/failover/raft.lua#L45-L56

PR #1953, opened 2022-10-14
Delete instances before box.cfg.replication change

### LINKED ISSUE #1948: Remove expelled instances from `box.space._cluster`
Expelled instances stay in `box.space._cluster`. We should clear them.

yngvar-antonsson:

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Close #1948

### DIFF
--- a/cartridge/failover/raft.lua
+++ b/cartridge/failover/raft.lua
@@ -47,10 +47,11 @@ local function get_appointments(topology_cfg)
             local member = membership.get_member(server.uri)
 
             if member ~= nil
-            and member.payload.leader_uuid ~= nil
+            and member.payload.raft_leader ~= nil
+            and member.payload.raft_leader ~= 0
             and member.payload.raft_term or 0 >= latest_term
             then
-                latest_leader = member.payload.leader_uuid
+                latest_leader = member.payload.raft_leader
                 latest_term = member.payload.raft_term or 0
             end
         end
@@ -74,6 +75,7 @@ local function on_election_trigger()
         membership.set_payload('leader_uuid', vars.leader_uuid)
     end
     membership.set_payload('raft_term', election.term)
+    membership.set_payload('raft_leader', leader.uuid)
 end
 
 _G.__cartridge_on_election_trigger = on_election_trigger
@@ -129,6 +131,7 @@ local function cfg()
     end
 
     membership.set_payload('raft_term', box.info.election.term)
+    membership.set_payload('raft_leader', box.info.election.leader)
 
     return true
 end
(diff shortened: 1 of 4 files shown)


filonenko-mikhail (approved):

cartridge/failover/raft.lua
```diff
@@ -66,9 +66,9 @@ local function on_election_trigger()
     local box_info = box.info()
     local election = box_info.election
 
-    local leader = box_info.replication[election.leader] or {}
+    local leader = box_info.replication[election.leader]
 
-    if vars.leader_uuid ~= leader.uuid then
+    if leader ~= nil and vars.leader_uuid ~= leader.uuid then
```
R-omk, line 71: If a check is added here for the presence of a elected leader exists, then the payload of raft term will be updated, but the payload leader will not be updated. This will cause an algorithm (in function get_appointments) that relies on the fact that a newer term contains up-to-date data will actually get outdated data in this case.

R-omk: https://github.com/tarantool/cartridge/issues/1900#issuecomment-1297085086

yngvar-antonsson: @R-omk Hi! I updated the code to check terms and leaders. Does it fix your problem?

cartridge/failover/raft.lua
```diff
@@ -129,6 +131,7 @@ local function cfg()
     end
 
     membership.set_payload('raft_term', box.info.election.term)
+    membership.set_payload('raft_leader', box.info.election.leader)
```
R-omk, line 134: `box.info.election.leader` is local number but need to store  uuid

R-omk: image

R-omk: Also box.info.election.leader may be zero  during start of new election process

R-omk: I'm not sure, but it seems there is no guarantee that the same number must correspond to the same uuid on different instances

cartridge/failover/raft.lua
R-omk: If by "problem" you mean issue1900 then I'm not sure, because leader_uuid is only set to member in a trigger that might never have been executed after the cartridge was started.

In terms of raft info consistency, it looks better, but it's still better to keep the payload as one indivisible object, and deal with the data type confusion that I wrote about in another comment.

cartridge/failover/raft.lua
yngvar-antonsson: > I'm not sure, but it seems there is no guarantee that the same number must correspond to the same uuid on different instances

On different instances box.info.election.leader is the same as numbers in box.space._cluster, so it corresponds to the same uuid if everything was ok in election process

vaintrub (approved):

yngvar-antonsson:
Will be reworked in #2081 and other follow-up PRs.

PR #1959, opened 2022-10-19
Extend authorization tests with new error message

drewdzzz:
Tarantool is going to replace an error thrown in case of entering an invalid password and entering the login of a non-existent user during authorization. The patch updates authorization tests correspondingly.


drewdzzz:
PR that updates the error message in Tarantool: https://github.com/tarantool/tarantool/pull/7827

yngvar-antonsson (approved):

PR #1972, opened 2022-11-07
Remove error stack

### LINKED ISSUE #1932: Do not send error stack to web ui
subj

filonenko-mikhail:

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Close #1932


cartridge/auth.lua
```diff
@@ -697,6 +701,7 @@ end
 local function init(httpd, opts)
     checks('table', {
         prefix = 'string',
+        disable_errstack = '?boolean',
```
yngvar-antonsson, line 704: you don't really use it in the code. `vars.disable_errstack` is always false

cartridge/graphql.lua
```diff
@@ -335,6 +339,9 @@ local function execute_graphql(req)
     local resp, err = e_graphql_internal:pcall(_execute_graphql, req)
     if resp == nil then
         log.error('%s', err)
+        if type(err) == 'table' and vars.disable_errstack then
```
yngvar-antonsson, line 342: The same here. `vars.disable_errstack` is always `false`. You didn't add that parameter in init function. I don't understand, why the tests are passed.

test/integration/graphql_test.lua
```diff
@@ -225,6 +225,21 @@ function g.test_error_extensions()
         extensions['io.tarantool.errors.class_name'],
         'Invalid cluster topology config'
     )
+
+    cluster.main_server:eval([[
+        require('cartridge.vars').new('cartridge.graphql').disable_errstack = true
+    ]])
```
yngvar-antonsson, line 231: That's why your tests passed. Don't use direct variable setting. Instead, look at optional flags in entrypoints and do the same way:
https://github.com/tarantool/cartridge/blob/9f00bdb2f636bd6225c304fec168b0ab5d4f542e/test/entrypoint/srv_basic.lua#L169-L172

yngvar-antonsson (changes requested):

cartridge/auth.lua
```diff
@@ -343,6 +344,9 @@ local function login(req)
         }
     elseif err ~= nil then
         log.error('%s', err)
+        if type(err) == 'table' and vars.disable_errstack then
```
yngvar-antonsson, line 347:
```suggestion
        if type(err) == 'table' and vars.disable_errstack == true then
```
Otherwise, you may have `box.NULL` in `vars.disable_errstack`

cartridge/graphql.lua
```diff
@@ -315,6 +316,9 @@ local function _execute_graphql(req)
         local extensions = err.graphql_extensions or {}
         extensions['io.tarantool.errors.class_name'] = err.class_name
         extensions['io.tarantool.errors.stack'] = err.stack
+        if vars.disable_errstack then
```
yngvar-antonsson, line 319:
```suggestion
        if vars.disable_errstack == true then
```

cartridge/graphql.lua
yngvar-antonsson, line 342:
```suggestion
        if type(err) == 'table' and vars.disable_errstack == true then
```

yngvar-antonsson (changes requested):

yngvar-antonsson (approved):

PR #1990, opened 2022-11-24
Warn users when they have invalid format

### LINKED ISSUE #1985: Produce a warning when spaces have invalid formats
## Motivation 

There is problem https://github.com/tarantool/tarantool/issues/7841

There is code fix: https://github.com/tarantool/tarantool/wiki/Fix-illegal-field-type-in-a-space-format-when-upgrading-to-2.10.4

## Task

The task is to adopt code fix to produce warning in cartridge.

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #1985

### DIFF
--- a/cartridge/confapplier.lua
+++ b/cartridge/confapplier.lua
@@ -28,6 +28,7 @@ local remote_control = require('cartridge.remote-control')
 local cluster_cookie = require('cartridge.cluster-cookie')
 local ClusterwideConfig = require('cartridge.clusterwide-config')
 local logging_whitelist = require('cartridge.logging_whitelist')
+local invalid_format = require('cartridge.invalid-format')
 
 yaml.cfg({
     encode_load_metatables = false,
@@ -376,6 +377,8 @@ local function boot_instance(clusterwide_config)
         'Unexpected state ' .. vars.state
     )
 
+    invalid_format.start_check()
+
     local topology_cfg = clusterwide_config:get_readonly('topology') or {}
     for _, server in pairs(topology_cfg.servers or {}) do
         if server ~= 'expelled' then
@@ -608,6 +611,8 @@ local function boot_instance(clusterwide_config)
             remote_control.resume()
         end
     end
+
+    invalid_format.end_check()
 
     -- Box is ready, start listening full-featured iproto protocol
     remote_control.stop()
--- a/cartridge/confapplier.lua
+++ b/cartridge/confapplier.lua
@@ -377,8 +377,6 @@ local function boot_instance(clusterwide_config)
         'Unexpected state ' .. vars.state
     )
 
-    invalid_format.start_check()
-
     local topology_cfg = clusterwide_config:get_readonly('topology') or {}
     for _, server in pairs(topology_cfg.servers or {}) do
         if server ~= 'expelled' then
@@ -545,11 +543,12 @@ local function boot_instance(clusterwide_config)
     -- This operation may be long
     -- It recovers snapshot
     -- Or bootstraps replication
+    invalid_format.start_check()
     local snap1 = hotreload.snap_fibers()
     box.cfg(box_opts)
     local snap2 = hotreload.snap_fibers()
     hotreload.whitelist_fibers(hotreload.diff(snap1, snap2))
-
+    invalid_format.end_check()
     require('membership.options').SUSPICIOUSNESS = true
 
     local username = cluster_cookie.username()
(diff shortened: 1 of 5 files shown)


cartridge/invalid-format.lua
```diff
@@ -0,0 +1,62 @@
+local log = require('log')
+
+local illegal_types = {
+    [''] = true,
+    ['n'] = true,
+    ['nu'] = true,
+    ['s'] = true,
+    ['st'] = true,
+}
```
filonenko-mikhail, line 9: very good

filonenko-mikhail (approved):

cartridge/invalid-format.lua
```diff
@@ -0,0 +1,68 @@
...
+local function before_replace(_, tuple)
+    if tuple == nil then return tuple end
+
+    local name = tuple[3]
+    local format = tuple[7]
+
+    for _, field_def in ipairs(format) do
+        if illegal_types[field_def.type] then
+            _G.__cartridge_invalid_format_spaces[name] = true
+        end
```
Totktonada, line 23: I would also look at `field_def[2]`. AFAIU a user may have a space created, when the old format representation was the default one.

cartridge/invalid-format.lua
```diff
@@ -0,0 +1,68 @@
...
+-- _space trigger.
+local function before_replace(_, tuple)
+    if tuple == nil then return tuple end
+
+    local name = tuple[3]
+    local format = tuple[7]
+
+    for _, field_def in ipairs(format) do
+        if illegal_types[field_def.type] then
+            _G.__cartridge_invalid_format_spaces[name] = true
```
Totktonada, line 22: Is it verified with `strict.on()` / on a Debug build? I see `rawset()` above, so existence of the global variable is not tracked and so assigning to it in the script mode will raise an error.

The easier way to solve it is to assign the global directly:

```lua
_G.__cartridge_invalid_format_spaces = {}
```

It is allowed on the module level.

The alternative is to make all accesseses via `rawget()`.

(However, please, don't lean on my words and verify it.)

BTW, why not use a module-local variable?

cartridge/invalid-format.lua
```diff
@@ -0,0 +1,68 @@
...
+            return
+        end
+        box.space._space:before_replace(nil, before_replace)
+        box.ctl.on_schema_init(nil, on_schema_init)
+
+        if next(_G.__cartridge_invalid_format_spaces) then
+            log.warn(
+                "You have spaces with invalid format: " ..
+                spaces_name_to_str() ..
+                ". Fix it before use Tarantool 2.10.4 - it's deprecated and " ..
```
Totktonada, line 60: BTW, the trigger allows to bootstrap on tarantool 2.10.4 as well.

cartridge/confapplier.lua
```diff
@@ -609,6 +612,8 @@ local function boot_instance(clusterwide_config)
         end
     end
 
+    invalid_format.end_check()
```
Totktonada, line 615: We definitely can exit from the function before this statement. Is it crucial?

cartridge/invalid-format.lua
```diff
@@ -0,0 +1,68 @@
...
+    end,
+    end_check = function()
+        if box.ctl.on_schema_init == nil then
+            return
+        end
+        box.space._space:before_replace(nil, before_replace)
+        box.ctl.on_schema_init(nil, on_schema_init)
+
+        if next(_G.__cartridge_invalid_format_spaces) then
+            log.warn(
```
Totktonada, line 57: Shouldn't we put it into issues that are shown in the web interface?

cartridge/invalid-format.lua
yngvar-antonsson: @Totktonada 
To the last question - I used it in tests only. So, I can move it under the test flag, that we've used before in our code. 
And all the tests run with `strict.on()`.

cartridge/invalid-format.lua
yngvar-antonsson: Cool idea!

Totktonada, line 57: The issue (#1985) asks to migrate the bad formats. Is there implecit agreement to issue a warning instead?

BTW, the issue will not be shown on tarantool 2.10.4. We'll get a fail on bootstrap. Just noted. Don't know, whether it is intentional.

cartridge/invalid-format.lua
Totktonada: Hm. Okay. Maybe I don't know something about the strict mode. If it works, disregard my comment.

Totktonada (approved):
I don't have enough experience to actually review cartridge's code. I left some comments and, I guess, I have nothing more to add. I'll approve, but, please, don't take my approve here with any level of trust.

PR #1994, opened 2022-12-05
Use require('log') instead of package.loaded.log in tests (#1993)

### LINKED ISSUE #1993: Use `require('log')` instead of `package.loaded.log` in tests
According to Lua 5.1 Reference Manual, there is no sense in accessing already loaded modules via `package.loaded` table, the generic interface `require()` does the same thing:

**[require (modname)](http://www.lua.org/manual/5.1/manual.html#pdf-require)**
> Loads the given module. The function starts by looking into the [package.loaded](http://www.lua.org/manual/5.1/manual.html#pdf-package.loaded) table to determine whether modname is already loaded. If it is, then require returns the value stored at package.loaded[modname]. Otherwise, it tries to find a loader for the module.

**[package.loaded](http://www.lua.org/manual/5.1/manual.html#pdf-package.loaded)**
> A table used by [require](http://www.lua.org/manual/5.1/manual.html#pdf-require) to control which modules are already loaded. When you require a module modname and package.loaded[modname] is not false, [require](http://www.lua.org/manual/5.1/manual.html#pdf-require) simply returns the value stored there.

Furthermore, the Tarantool new feature https://github.com/tarantool/tarantool/issues/3211 won't work well if `log` object is changed directly in `package.loaded.log`, rather than using `require('log')`.

Gumix:
According to Lua 5.1 Reference Manual, there is no sense in accessing already loaded modules via `package.loaded` table, the generic interface `require()` does the same thing:

**[require (modname)](http://www.lua.org/manual/5.1/manual.html#pdf-require)**
> Loads the given module. The function starts by looking into the [package.loaded](http://www.lua.org/manual/5.1/manual.html#pdf-package.loaded) table to determine whether modname is already loaded. If it is, then require returns the value stored at package.loaded[modname]. Otherwise, it tries to find a loader for the module.

**[package.loaded](http://www.lua.org/manual/5.1/manual.html#pdf-package.loaded)**
> A table used by [require](http://www.lua.org/manual/5.1/manual.html#pdf-require) to control which modules are already loaded. When you require a module modname and package.loaded[modname] is not false, [require](http://www.lua.org/manual/5.1/manual.html#pdf-require) simply returns the value stored there.

Furthermore, the Tarantool new feature https://github.com/tarantool/tarantool/issues/3211 won't work well if `log` object is changed directly in `package.loaded.log`, rather than using `require('log')`.

Close #1993


test/integration/logger_test.lua
```diff
@@ -35,8 +35,8 @@ function g.test_cfg()
 
     print()
     g.server:start()
-    g.server:call('package.loaded.log.info', {'Info message'})
-    g.server:call('package.loaded.log.error', {'Error message'})
+    g.server:call('require('log').info', {'Info message'})
```
yngvar-antonsson, line 38: This won't work. 
1) You can't call a function inside net.box:call 
2) This is a syntax error

I suggest leaving those calls as is because there is no hack with changing package functions.

yngvar-antonsson (changes requested):

yngvar-antonsson:
Thank you for the patch! Left one comment. 
And please run luacheck and luatest and put screenshots in the comments.

test/integration/logger_test.lua
Gumix: OK, reverted changes in `logger_test.lua`.

yngvar-antonsson (approved):

Issue #1998, opened 2022-12-08
Do not override logging functions in tests


Gumix:
There are 2 tests ([unit/upload_test.lua](https://github.com/tarantool/cartridge/blob/master/test/unit/upload_test.lua#L18) and [integration/twophase_test.lua](https://github.com/tarantool/cartridge/blob/master/test/integration/twophase_test.lua#L58)) that override logging functions in the `log` module.
Unfortunately this hack doesn't work with the new Tarantool feature «[Per-module logging](https://github.com/tarantool/tarantool/issues/3211)».

The [first attempt](https://github.com/tarantool/cartridge/pull/1994) to fix it didn't actually help, because each Lua module will contain an own instance of the logger, but the logging function is overwritten only in a single module (e.g. `unit/upload_test.lua`), while other modules (e.g. `cartridge/upload.lua`) will use original functions from the log module.

It's possible to fix integration/twophase_test.lua by overriding the `require` function for all modules:
<details><summary>Here is the patch</summary>

```diff
diff --git a/test/entrypoint/srv_loghack.lua b/test/entrypoint/srv_loghack.lua
new file mode 100755
index 00000000..21d7ee01
--- /dev/null
+++ b/test/entrypoint/srv_loghack.lua
@@ -0,0 +1,24 @@
+#!/usr/bin/env tarantool
+
+local require_original = require
+_G.require = function(name)
+    local module = require_original(name)
+    if name == 'log' then
+        module.warn = function(...)
+            if rawget(_G, '__log_warn') == nil then
+                rawset(_G, '__log_warn', {})
+            end
+            table.insert(_G.__log_warn, string.format(...))
+        end
+        module.error = function(...)
+            if rawget(_G, '__log_error') == nil then
+                rawset(_G, '__log_error', {})
+            end
+            table.insert(_G.__log_error, string.format(...))
+        end
+    end
+    return module
+end
+
+local helpers = require('test.helper')
+dofile(helpers.entrypoint('srv_basic'))
diff --git a/test/integration/twophase_test.lua b/test/integration/twophase_test.lua
index 05a66545..b3c05a3f 100644
--- a/test/integration/twophase_test.lua
+++ b/test/integration/twophase_test.lua
@@ -33,7 +33,7 @@ end
 g.before_all(function()
     g.cluster = helpers.Cluster:new({
         datadir = fio.tempdir(),
-        server_command = helpers.entrypoint('srv_basic'),
[...]

PR #1999, opened 2022-12-08
Temporary XFAIL tests that override logging functions

Gumix:
See #1998 for details.

luacheck and luatest passed.

### DIFF
--- a/test/integration/twophase_test.lua
+++ b/test/integration/twophase_test.lua
@@ -1,4 +1,5 @@
 local fio = require('fio')
+local log = require('log')
 
 local t = require('luatest')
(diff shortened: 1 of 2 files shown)


yngvar-antonsson:
Thank you for your patch!

It won't work.
You need to to use `xfail_if(version > ...)`
Снимок экрана 2022-12-09 в 11 38 38

Gumix:
> You need to to use xfail_if(version > ...)

Fixed, now they are XFAILed only if "per-module logging" is enabled.

yngvar-antonsson (approved):

PR #2016, opened 2023-01-10
Use default values of `connect_quorum` and `sync_timeout` in first `box.cfg` call

yngvar-antonsson:
I didn't forget about

- [x] Changelog


yngvar-antonsson:
@sergepetrenko, thank you for helping with the patch!

sergepetrenko (approved):
Thanks for working on this!
LGTM.

PR #2047, opened 2023-03-06
Metrics embedding updates

DifferentialOrange:
Add `cartridge.roles.metrics` from tarantool/metrics [1], as well as utility code and integration tests.

metrics will become a part of the core tarantool after 2.11.1.

Since cartridge is not a part of the core tarantool, we cannot embed role together with module code. Core test engine also do not support rocks integration tests, so we must place them here.

On the other hand, due to backward compatibility we do not plan to remove the code from tarantool/metrics itself, so, for some time, there would be two implementations of the role and tests. Role should work fine with all metrics version with role support (0.3.0 and newer). (Some minor behavior may differ for old versions: for example, this role does not enable cartridge metrics by default before 0.10.0.)

I didn't forget about

- [x] Tests
- [x] Changelog
- Documentation

Part of tarantool/tarantool#7725


.github/workflows/backend-test.yml
```diff
@@ -88,9 +88,26 @@ jobs:
...
+            metrics: '0.13.0'
+          - tarantool: '2.10'
+            etcd: 'v2.3.8'
+            metrics: '0.14.0'
+          - tarantool: '2.10'
+            etcd: 'v2.3.8'
+            metrics: '0.15.1'
+          - tarantool: '2.10'
+            etcd: 'v2.3.8'
+            metrics: '0.16.0'
```
yngvar-antonsson, line 110: I propose to leave only the latest metrics version. Tests in cartridge are hard-weighted, so several additional workflows could be bad for PR performance.

yngvar-antonsson: As we won't remove the code from the metrics repo, maybe we should add it as a submodule?

yngvar-antonsson: And maybe we should add metrics to the rockspec as a dependency?

DifferentialOrange: > And maybe we should add metrics to the rockspec as a dependency?

But cartridge works fine without metrics, unless you decide to enable `cartridge.roles.metrics` with no `metrics`.

DifferentialOrange: > But I have heard that building a Lua module with submodules is rather painful.

Or maybe not. I'll try this.

PR #2076, opened 2023-04-18
Add maentenance note

filonenko-mikhail:

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Close #???

### DIFF
--- a/rst/troubleshooting.rst
+++ b/rst/troubleshooting.rst
@@ -435,12 +435,12 @@ Or you can use the following workaround:
 A `Connection refused` error in unconfigured instance since 2.7.4
 -----------------------------------------------------------------
 
-Before v2.7.4, an unconfigured instance was bound to the `0.0.0.0` interface.
+Before `v2.7.4 <https://github.com/tarantool/cartridge/blob/master/CHANGELOG.rst#274---2022-04-11>`_, an unconfigured instance was bound to the `0.0.0.0` interface.
 Given that `0.0.0.0` accepts connections on any IP address assigned to the machine,
 this might impose additional security risks.
 
-With v2.7.4 version, an unconfigured instance resolves the `advertise_uri` host and binds to it.
-You can check that the instance's `advertise_uri` is resolved to a network interface (not loopback) as follows:
+With `v2.7.4 <https://github.com/tarantool/cartridge/blob/master/CHANGELOG.rst#274---2022-04-11>`_ version, an unconfigured instance resolves the `advertise_uri` host and binds to it.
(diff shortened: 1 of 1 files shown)


rst/troubleshooting.rst
```diff
@@ -432,3 +432,16 @@ Or you can use the following workaround:
...
+Unconfigured instance binds to `0.0.0.0` interface before 2.7.4 .
+
+Since 2.7.4 unconfigure instance resolves advertise uri host and binds to it.
+
+Check that instance advertise uri resolved to network interface (not loopback).
+
+.. code-block:: bash
+
+    dig +short place_advertise_uri_here
+
```
andreyaksenov, line 447: I suggest the following wording:

```
A `Connection refused` error in unconfigured instance since 2.7.4
-----------------------------------------------------------------

Before v2.7.4, an unconfigured instance was bound to the `0.0.0.0` interface.
Given that `0.0.0.0` accepts connections on any IP address assigned to the machine,
this might impose additional security risks.

With v2.7.4 version, an unconfigured instance resolves the `advertise_uri` host and binds to it.
You can check that the instance's `advertise_uri` is resolved to a network interface (not loopback) as follows:

.. code-block:: bash
    dig +short place_advertise_uri_here
```

Note that I've also added information about the reason for this change (security). It's a good practice to describe behaviour/breaking changes in the following way:
- What Changed
- Reasons for Change
- Impact on Existing Apps
- How to Update Existing Apps (optional)
- How to Revert to Previous Behavior (optional)

Maybe, we can use this layout in all similar tickets, so a user can understand our reasoning and consequences.

andreyaksenov (changes requested):

rst/troubleshooting.rst
filonenko-mikhail: thnks a lot

andreyaksenov: sorry, forgot to add two links. Could you please replace the version and `advertise_uri` with corresponding links?
```
- v2.7.4 -> [v2.7.4](https://github.com/tarantool/cartridge/blob/master/CHANGELOG.rst#274---2022-04-11)
- `advertise_uri` -> [advertise_uri](https://www.tarantool.io/en/doc/latest/book/cartridge/cartridge_dev/#configuration-basics)
```

andreyaksenov: sorry, forgot to add two links. Could you please replace the version and `advertise_uri` with corresponding links?
```
- v2.7.4 -> `v2.7.4 <https://github.com/tarantool/cartridge/blob/master/CHANGELOG.rst#274---2022-04-11>`_
- `advertise_uri` -> `advertise_uri <https://www.tarantool.io/en/doc/latest/book/cartridge/cartridge_dev/#configuration-basics>`_
```

PR #2098, opened 2023-05-12
Custom health handler + refactoring

### LINKED ISSUE #2097: Custom health handler
# Description of the problem
To get a custom response format for a health test, a hack is used:
```lua
local cartridge_health = require('cartridge.health')

local cartridge_health_func = cartridge_health.is_healthy
cartridge_health.is_healthy = function (req)
   local result_orig = cartridge_health_func()
   local healthy = type(result_orig) == 'table' and result_orig.status == 200 and result_orig.body:endswith('is OK')
      
   local resp = req:render{
      json = {
         app_name = instance_params.app_name,
         instance_name = instance_params.instance_name or instance_params.alias or '',
         instance_health = healthy,
      }
   }
   resp.status = 200
   return resp
end
```

In order to avoid a hack, if you need to give a response about health in a custom format, you must:
1. Make your own health handler. To make your own handler, you need to implement the logic for extracting the handler address from the cluster configuration. The very function of determining health can be used by the existing one. But...
2. The cartridge.health.is_healthy function returns a non-boolean value. Although the name suggests it. It returns a structure with status and text. What if the answer is changed to this: `{status = 200, body = 'Instance is dead'}`? Logically, this is not an error. I would like a clear answer - yes or no.
3. You need to turn off the same handle for Cartridge. And this increases the volume of a custom solution.

# Suggested solution
Instead, I propose to give the user the function of setting their own handler to check health.

a1div0:
See problem description [here](https://github.com/tarantool/cartridge/issues/2097)
I propose to give the user the function of setting their own handler to check health.

``` lua
local metrics = require('cartridge.roles.metrics')
metrics.set_custom_is_health_handler(function(req)
    local health = require('cartridge.health')
    local resp = req:render{
        json = {
            my_healthcheck_format = health.is_healthy()
        }
    }
    resp.status = 200
    return resp
end)
```

# Changes
- `metrics.lua` - added function `set_custom_health_handler`, subj
- `health.lua` - divided one function `is_healthy` to three: `is_healthy`, `is_healthy_handler`, `member_is_healthy` - the content of each of which is consistent with the title
- `rpc.lua` - We have a code repetition, and a text link to the original piece is made from the health module. So, replaced code to use `health.member_is_healthy`, and rename `member_is_healthy` to `instance_is_healthy`, because he argument is instance

# Checklist
- [x] Tests
- [x] Changelog
- [x] Documentation

# Links
Close #2097

### DIFF
--- a/cartridge/roles/metrics.lua
+++ b/cartridge/roles/metrics.lua
@@ -8,6 +8,7 @@ local metrics = require('metrics')
 -- Backward compatibility with metrics 0.16.0 or older.
 local metrics_stash_supported, metrics_stash = pcall(require, 'metrics.stash')
 local log = require('log')
+local health = require('cartridge.health')
 
 local metrics_vars = require('cartridge.vars').new('metrics_vars')
 metrics_vars:new('current_paths', {})
@@ -25,10 +26,7 @@ local handlers = {
         local http_handler = require('metrics.plugins.prometheus').collect_http
         return http_handler(...)
     end,
-    ['health'] = function(...)
-        local http_handler = require('cartridge.health').is_healthy
-        return http_handler(...)
-    end,
+    ['health'] = health.default_is_healthy_handler,
 }
 
 local function set_labels(custom_labels)
@@ -246,6 +244,17 @@ local function stop()
     metrics_vars.custom_labels = {}
 end
 
+local function set_is_health_handler(new_handler)
+    handlers['health'] = new_handler or health.default_is_healthy_handler
+
+    local paths = table.copy(metrics_vars.current_paths)
+    for path, _ in pairs(metrics_vars.current_paths) do
+        metrics_vars.current_paths[path] = ''
+    end
+
+    apply_routes(paths)
+end
+
 return setmetatable({
     role_name = 'metrics',
     permanent = true,
@@ -255,4 +264,5 @@ return setmetatable({
     apply_config = apply_config,
     set_export = set_export,
     set_default_labels = set_default_labels,
+    set_is_health_handler = set_is_health_handler,
 }, { __index = metrics })
--- a/test/integration/metrics_health_test.lua
+++ b/test/integration/metrics_health_test.lua
@@ -51,11 +51,45 @@ g.after_each(function()
     end)
 end)
 
+g.after_test('test_metrics_custom_is_health_handler', function()
+    local main_server = g.cluster:server('main')
+    main_server:exec(function()
+        local cartridge = require('cartridge')
+        local metrics = cartridge.service_get('metrics')
+        metrics.set_is_health_handler(nil)
+    end)
+end)
+
 g.test_metrics_health_handler = function()
     helpers.upload_default_metrics_config(g.cluster)
     local main_server = g.cluster:server('main')
     local resp = main_server:http_request('get', '/health', {raise = false})
     t.assert_equals(resp.status, 200)
+    t.assert_equals(resp.body, 'main is OK')
+end
+
+g.test_metrics_custom_is_health_handler = function()
+    local main_server = g.cluster:server('main')
+    main_server:exec(function()
+        local cartridge = require('cartridge')
+        local metrics = cartridge.service_get('metrics')
+        metrics.set_is_health_handler(function(req)
+            local health = require('cartridge.health')
+            local resp = req:render{
+                json = {
+                    my_healthcheck_format = health.is_healthy()
+                }
+            }
+            resp.status = 200
+            return resp
+        end)
+    end)
+
+    helpers.upload_default_metrics_config(g.cluster)
+    local resp = main_server:http_request('get', '/health', {raise = false})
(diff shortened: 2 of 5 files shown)


test/integration/metrics_health_test.lua
```diff
@@ -35,6 +35,14 @@ g.after_all(function()
     fio.rmtree(g.cluster.datadir)
 end)
 
+g.before_each(function()
+    local main_server = g.cluster:server('main')
+    main_server:exec(function()
+        local metrics = require('cartridge.roles.metrics')
+        metrics.set_custom_is_health_handler(nil)
```
DifferentialOrange, line 42: Maybe it's better to make a single cleanup in `g.after_test('test_metrics_custom_is_health_handler', function)`

test/integration/metrics_health_test.lua
```diff
@@ -58,6 +66,29 @@ g.test_metrics_health_handler = function()
...
+                json = {
+                    my_healthcheck_format = health.is_healthy()
+                }
+            }
+            resp.status = 200
+            return resp
+        end)
+    end)
+
+    helpers.upload_default_metrics_config(g.cluster)
```
DifferentialOrange, line 85: Should this one work if I switch handle setup and configuration update?

DifferentialOrange (commented):
Thank you for you patch! See my comments below.

I have raised the issue in DMs with both you and @yngvar-antonsson , but it's better to replicate it here so it would be clear. For now, the metrics cartridge role is in the state of migration. The reason of this migration is the migration of tarantool/metrics package to tarantool/tarantool (and its cartridge components, `cartridge.roles.metrics` and `cartridge.health`, to tarantool/cartridge). But the core release 2.11.1 with embedded metrics is expected to appear not earlier than two-three months from now (and we'll also need to wait until people migrate to it). metrics role code is not yet removed from the metrics. So by installing metrics+cartridge user will get files either from metrics or from cartridge. Users with old metrics+new cartridge will not get rid of this behavior even after code would be removed from the tarantool/metrics.

So if we need ro add new API, we should
1. Add it to tarantool/metrics too or remove the code from the repo
2. Ensure that user of new API uses both new cartridge and new metrics.
Maybe we'll need to workaround something in the code.

test/integration/metrics_health_test.lua
```diff
@@ -58,6 +66,29 @@ g.test_metrics_health_handler = function()
     t.assert_equals(resp.status, 200)
 end
 
+g.test_metrics_custom_is_health_handler = function()
+    local main_server = g.cluster:server('main')
+    main_server:exec(function()
+        local metrics = require('cartridge.roles.metrics')
```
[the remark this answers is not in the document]
a1div0: Replaced.
What is the difference
```
require('cartridge.roles.metrics')
```
and
```
cartridge.service_get('metrics')
```
?

test/integration/metrics_health_test.lua
a1div0: Yes. `set_is_health_handler` and `upload_default_metrics_config` work independently.

cartridge/roles/metrics.lua
```diff
@@ -246,6 +244,17 @@ local function stop()
     metrics_vars.custom_labels = {}
 end
 
+local function set_custom_is_health_handler(new_handler)
+    handlers['health'] = new_handler or health.default_is_healthy_handler
+
+    local paths = table.copy(metrics_vars.current_paths)
+    for path, _ in pairs(metrics_vars.current_paths) do
+        metrics_vars.current_paths[path] = '_del_'
```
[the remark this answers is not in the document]
a1div0: This is to force a route update. Indeed, an empty string can be used.
Something I got excited with `_del_` )

test/integration/metrics_health_test.lua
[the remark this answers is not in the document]
DifferentialOrange: `service_get` is by-design way to interact with role enabled on the instance. The difference between `require` and `get_service` is that `require` is always available when the file exist, but `get_service` returns non-nil module only if role is actually enabled. It doesn't that much of an issue in case of a permanent role `cartridge.roles.metrics`, especially in tests, but since it is common to copypaste test contents to applications, it is better to call appropriate API, if possible.

cartridge/roles/metrics.lua
```diff
@@ -255,4 +264,5 @@ return setmetatable({
     apply_config = apply_config,
     set_export = set_export,
     set_default_labels = set_default_labels,
+    set_is_health_handler = set_is_health_handler,
```
DifferentialOrange, line 267: Since this section is called `'health'` in clusterwide config, I think it's fine to call it just `set_health_handler` here.

DifferentialOrange (commented):
Thank you for the updates! Should be fine from my side after resolving the remaining comments.

yngvar-antonsson:
Thank you for the patch! `cartridge.roles.metrics` moved to tarantool/cartridge-metrics-role, so please reopen the PR in its repo.

Issue #1361, opened 2021-04-01
Logging under systemd


Kasen:
Journald may not expose `/dev/log` device, which is a compatibility layer with syslog. But cartridge detects systemd and configure tarantool for writing to syslog. But if `/dev/log` is not exposed, Tarantool doesn't start with error:
```
Apr 01 10:40:01 **** systemd[1]: Started Tarantool service: consumer.****.
Apr 01 10:40:01 **** tarantool[28583]: SystemError No such file or directory: No such file or directory
Apr 01 10:40:01 **** tarantool[28583]: SystemError syslog logger: No such file or directory: No such file or directory
Apr 01 10:40:01 **** tarantool[28583]: failed to initialize logging subsystem
Apr 01 10:40:01 **** systemd[1]: consumer.****.service: main process exited, code=exited, status=1/FAILURE
```

How to reproduce on centos 7:
1. systemctl stop systemd-journald.socket
2. systemctl start crtg-service

I suggest checking that `/dev/log` exists and accepts connections.

PR #2118, opened 2023-07-19
cfg: fix syslog availability check

grafin:
Historically log defaulted to syslog, only if the instance was started in systemd environment. Sometime after additional check was added because of #1361. New check, checks if syslog UNIX socket is available. No need to check for systemd if we already check that syslog socket exists and even functions.

First commit removes the systemd check from log initialization because:
1) It is kinda hacky (and doesn't work on Centos 7.9 and RHEL 7.9)
2) It is not needed

Second commit tries to treat `/dev/log` as a UDP socket (SOCK_DGRAM), again, on RHEL 7.9 and Centos 7.9 it is that way, and simple `socket.connect` returns "Protocol wrong type for socket" error. But syslog works fine.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,8 @@ Changed
 
 - Update ``cartridge-metrics-role`` dependency to `0.1.1 <https://github.com/tarantool/cartridge-metrics-role/releases/tag/0.1.1>`_.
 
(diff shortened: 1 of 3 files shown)


yngvar-antonsson (approved):

yngvar-antonsson:
Please add Changelog and merge!

PR #2123, opened 2023-07-24
test: send logs to stderr in tests

grafin:
After 5d60dfa86c08 logs are being sent syslog during tests. This makes it hard to access them. Lets return old behaviour during tests.

### DIFF
--- a/test/entrypoint/srv_basic.lua
+++ b/test/entrypoint/srv_basic.lua
@@ -203,6 +203,9 @@ local ok, err = errors.pcall('CartridgeCfgError', cartridge.cfg, {
     -- which doesn't support it yet.
     upload_prefix = package.loaded['cartridge.upload'] and '../upload',
     disable_errstack = disable_errstack,
+},
+{
+    log = '',
 })
 if not ok then
     log.error('%s', err)
--- a/test/entrypoint/srv_multisharding.lua
+++ b/test/entrypoint/srv_multisharding.lua
@@ -24,6 +24,9 @@ local ok, err = cartridge.cfg({
         'cartridge.roles.vshard-storage',
         'cartridge.roles.vshard-router',
     },
+},
+{
+    log = '',
 })
 
 if not ok then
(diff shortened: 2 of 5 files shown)


test/entrypoint/srv_stateboard.lua
```diff
@@ -3,3 +3,6 @@
 require('strict').on()
 local stateboard = require('cartridge.stateboard')
 stateboard.cfg()
+
+local ok, err = require('cartridge').cfg({}, {log = 'stderr'})
+assert(ok, err)
```
olegrok, line 8: Stateboard shouldn't call cartridge.cfg.

test/entrypoint/srv_empty.lua
```diff
@@ -9,6 +9,9 @@ local ok, err = remote_control.bind(
 )
 assert(ok, err)
 
+ok, err = require('cartridge').cfg({}, {log = 'stderr'})
+assert(ok, err)
```
olegrok, line 13: Is it possible to configure it with `log.cfg/box.cfg`. I'm not sure that direct cartridge.cfg call will work as expected.

`cartridge.cfg` doesn't work in the same way as `box.cfg`. While two sequential `box.cfg{}` can reload different parts of configuration (e.g. `replication` and `log`) `cartridge.cfg` updates whole configuration.

olegrok (commented):
Thanks for your patch. See my comments below

test/entrypoint/srv_stateboard.lua
grafin: Fixed.

test/entrypoint/srv_empty.lua
grafin: log.cfg can only be called once (it reports an error if called more then once trying to set `log` option). And `cartridge.cfg` calls it). There are two solutions to that:
1) Add the `log` option only to existent `cartridge.cfg` calls (currently implemented)
2) Set env variable `TARANTOOL_LOG=""`

I don't know which solution I like more, whats your opinion on that?

olegrok: Well, let's stay as is since it works

olegrok (approved):

yngvar-antonsson (approved):

PR #2170, opened 2023-12-13
Fixed invalid config parsing in argparse

### LINKED ISSUE #2169: When misconfigured `cartridge.cfg` fails with dubious error during startup
During yet-another-cluster-startup I've ran into an issue when cartridge produced a strange error executing `cartridge.cfg` function:

```
LuajitError: ...hov/Workspace/tarantool/cartridge/cartridge/argparse.lua:357: bad argument #1 to 'pairs' (table expected, got string)
fatal error, exiting the event loop
```

During a brief troubleshooting, I was able to make the shortest reproducer. The problem is within `cartridge.argparse` module. Here is the repro:

```
$ echo "default: option" > cfg.yml
$ TARANTOOL_CFG=cfg.yml tarantool -e "require('cartridge.argparse').parse()"
LuajitError: ...hov/Workspace/tarantool/cartridge/cartridge/argparse.lua:357: bad argument #1 to 'pairs' (table expected, got string)
fatal error, exiting the event loop
```

It turned out, that `argparse` module can load not only a single file, but a whole directory, if specified, in order to get its arguments. But sometimes it is possible for an ops team to accidentally (or intentionally) put inside this directory a config from a different service, which accepts key-value(string) configuration.

In this scenario I expect `argparse.parse()` either to get only its own options from a config directory tree, or at least explicitly say that it encountered invalid configuration such as `validateArgsSchemaError` (better pointing at the file and section location) and not fail with "mysterious" error.

I've reproduced it on the latest cartridge version at `master` branch:

```
$ git describe
2.8.4-13-gbe208ad6
```

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #2169

### DIFF
--- a/cartridge/argparse.lua
+++ b/cartridge/argparse.lua
@@ -41,6 +41,9 @@
 -- In this case, all files in the directory are parsed.
 -- To avoid conflicts, the same section mustn't repeat across different files.
 --
+-- Argparse logs if there were some ignored sections in configuration files.
+-- All sections should be key-value pairs as in example, otherwise you'll get an error.
+--
 -- @module cartridge.argparse
 
 local fio = require('fio')
@@ -48,6 +51,7 @@ local yaml = require('yaml')
 local json = require('json')
 local checks = require('checks')
 local errors = require('errors')
+local log = require('log')
 -- local base_dir = fio.abspath(fio.dirname(arg[0]))
 
 local vars = require('cartridge.vars').new('cartridge.argparse')
@@ -279,7 +283,13 @@ local function load_dir(dirname)
                     section_name, dirname, origin[section_name], fio.basename(f)
                 )
             end
-
+            if type(content) ~= 'table' then
+                return nil, ParseConfigError:new(
+                    'invalid content found in file %s in section %s: '..
+                    'expected a table of key-value pairs, got %s',
+                    fio.basename(f), section_name, type(content)
+                )
+            end
             for argname, argvalue in pairs(content) do
                 ret[section_name][argname:lower()] = argvalue
             end
@@ -351,12 +361,33 @@ local function parse_file(filename, search_name)
     end
 
     local ret = {}
+    local sections = {}
+    for section_name, _ in pairs(file_sections) do
+        sections[section_name] = true
+    end
     for _, section_name in ipairs(section_names) do
         local content = file_sections[section_name] or {}
-
+        if type(content) ~= 'table' then
+            return nil, ParseConfigError:new(
+                'invalid content found in file %s in section %s: '..
+                'expected a table of key-value pairs, got %s',
+                filename, section_name, type(content)
+            )
+        end
+        sections[section_name] = false
         for argname, argvalue in pairs(content) do
             ret[argname:lower()] = argvalue
         end
+    end
+    local names = {}
+    for section_name, ignored in pairs(sections) do
+        if ignored then
+           table.insert(names, section_name)
+        end
+    end
+    if #names > 0 then
+        log.warn('Some sections in config file %s were ignored: %s',
+            filename, table.concat(names, ', '))
     end
 
     return ret
--- a/cartridge/argparse.lua
+++ b/cartridge/argparse.lua
@@ -284,15 +284,17 @@ local function load_dir(dirname)
                 )
             end
             if type(content) ~= 'table' then
-                return nil, ParseConfigError:new(
-                    'invalid content found in file %s in section %s: '..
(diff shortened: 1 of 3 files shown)


cartridge/argparse.lua
```diff
@@ -353,6 +353,10 @@ local function parse_file(filename, search_name)
     local ret = {}
     for _, section_name in ipairs(section_names) do
         local content = file_sections[section_name] or {}
+        if type(content) ~= 'table' then
+            return nil, ParseConfigError:new('Invalid content: options should be '..
+                'in format `name: value`')
```
RunsFor, line 358: I would elaborate on the error message here. Here is some thoughts:
- it is more informative to say that a content must be a map or table with the format of `key: value`
- also it would be point out in which file and section the error occured

For example, for my reproducer the error message could look like this:

```
Invalid content found in file "cfg.yml" in section "default": expected a table of key-value pairs, got string
```

RunsFor (changes requested):
Thanks for a fast reaction!

But could you please be more informative in the error message?

Also, I've digged a little more into the issue and found another drawback with the relative code. Maybe you would decide to fix it right in place? Here is the link to my [comment](https://github.com/tarantool/cartridge/issues/2169#issuecomment-1854054179).

And if you don't mind, would you add any description about the PR? The checklist says, the tests are done. Was it manual test? Maybe it is better to add unit tests for a sake of possible future regression? What about documentation? Maybe it deserves to be mentioned explicitly, how config parsing works? What do you think?

yngvar-antonsson:
I propose:
1) make error messages more clear
2) not to brake cartridge if the user has config with ignored sections, but warn them in logs.

cartridge/argparse.lua
```diff
@@ -279,7 +283,13 @@ local function load_dir(dirname)
...
                 )
             end
-
+            if type(content) ~= 'table' then
+                return nil, ParseConfigError:new(
+                    'invalid content found in file %s in section %s: '..
+                    'expected a table of key-value pairs, got %s',
+                    fio.basename(f), section_name, type(content)
+                )
+            end
```
RunsFor, line 292: I believe this error is not necessary here. As I understand this code, we are loading directory into a flat config meaning each section has its own config.

But in some cases we parse sections, which are not needed actually. We can only figure it out later in the `parse_file` function, when we filter out all unnecessary sections.

But what to do here, when we do not have all the information which sections we needed? I suggest to just ignore invalid sections here, because what is a section?

Section is something with a name and a key-value map:

```yaml
section_name:
  key1: value1
  key2: value2
  # ...
```

But something like `section_name: a_string` is not actionally a section by definition. Why not just ignore it here?

RunsFor (commented):
Thanks for the improvements!

I've checked updated version locally and there is one scenario left I would like to discuss more. It is when you are loading directory, not file:

```bash
$ TARANTOOL_CFG=cfg tarantool -e "print(require('cartridge.argparse').parse())"
nil	ParseConfigError: invalid content found in file python.yml in section python: expected a table of key-value pairs, got string
stack traceback:
	...hov/Workspace/tarantool/cartridge/cartridge/argparse.lua:287: in function 'load_dir'
	...hov/Workspace/tarantool/cartridge/cartridge/argparse.lua:347: in function 'parse_file'
	...hov/Workspace/tarantool/cartridge/cartridge/argparse.lua:450: in function '_parse'
	...hov/Workspace/tarantool/cartridge/cartridge/argparse.lua:465: in function 'parse'
	(command line):1: in main chunk
```

I believe, we can just ignore invalid sections here. What do you think?

cartridge/argparse.lua
yngvar-antonsson: Ok, let's ignore everything.

RunsFor (approved):
LGTM

Thanks for your effort and patience!


This is my one-file check:

```bash
$ TARANTOOL_CFG=cfg.yml tarantool
Tarantool 2.11.1-0-g96877bd
type 'help' for interactive help
tarantool> require("yaml").decode(require("cartridge.utils").file_read(os.getenv("TARANTOOL_CFG")))
---
- {'cartridge': 'option', 'python': 'option', 'default': {'c': 'd'}}
...

tarantool> require('cartridge.argparse').parse()
invalid content ignored in file cfg.yml in section cartridge: expected a table of key-value pairs, got string
Some sections in config file cfg.yml were ignored: cartridge, python
---
- cfg: cfg.yml
  app_name: cartridge
  c: d
...
```

PR #2189, opened 2024-02-09
Fix raft failover state transitions

### LINKED ISSUE #2188: Master discovery with raft failover may return a dead node as leader
Reproduced by Tarantool integration tests: https://github.com/tarantool/tarantool/actions/runs/7829459994/job/21362395156

The test `integration.raft.test_kill_master` flaky fails with the following output:
```
Failed tests:
-------------

1) integration.raft.test_kill_master
.../work/tarantool/tarantool/test/integration/raft_test.lua:324: expected 
{
    ["bbbbbbbb-0000-0000-0000-000000000000"] = {
        ["bbbbbbbb-bbbb-0000-0000-000000000001"] = {master = true},
        ["bbbbbbbb-bbbb-0000-0000-000000000002"] = {master = false},
        ["bbbbbbbb-bbbb-0000-0000-000000000003"] = {master = false},
    },
    ["cccccccc-0000-0000-0000-000000000000"] = {["cccccccc-cccc-0000-0000-000000000001"] = {master = true}},
} to cover 
{
    ["bbbbbbbb-0000-0000-0000-000000000000"] = {
        ["bbbbbbbb-bbbb-0000-0000-000000000001"] = {master = false},
        ["bbbbbbbb-bbbb-0000-0000-000000000002"] = {master = true},
        ["bbbbbbbb-bbbb-0000-0000-000000000003"] = {master = false},
    },
}
```

However at https://github.com/tarantool/cartridge/blob/bb0cb61ac6d8dd336bd82be8ef598e3253419615/test/integration/raft_test.lua#L324-L329 `storage-3` is the only alive instance. Both `storage-1` and `storage-2` are stopped, and `storage-2` is the last known leader, so the test expects it to have `master = true` in the table.

Test logs at the relevant moment contain (remember that storage-1 is dead, so it's unclear how it could be appointed the new leader):
```
router-1 | 2024-02-08 12:49:39.003 [9925] main/144/vshard.failover._static_router/vshard.router I> New replica localhost:13304(admin@localhost:13304) for replicaset(id="bbbbbbbb-0000-0000-0000-000000000000", master=localhost:13303(admin@localhost:13303))
router-1 | 2024-02-08 12:49:39.004 [9925] main/144/vshard.failover._static_router/vshard.router I> Failovering step is finished. Schedule next after 1.000000 seconds
router-1 | 2024-02-08 12:49:39.016 [9925] main/173/cartridge.raft-failover/cartridge.failover I> Replicaset bbbbbbbb-0000-0000-0000-000000000000: new leader bbbbbbbb-bbbb-0000-0000-000000000001 ("localhost:13302"), was bbbbbbbb-bbbb-0000-0000-000000000002 ("localhost:13303")
```

Full failed test log: [job.log](https://github.com/tarantool/cartridge/files/14221461/job.log)

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog

Close #2188


filonenko-mikhail (approved):

cartridge/failover/raft.lua
```diff
@@ -69,10 +69,13 @@ local function on_election_trigger()
...
     if vars.leader_uuid ~= leader.uuid then
-        vars.leader_uuid = leader.uuid
-        vars.cache.is_leader = vars.leader_uuid == vars.instance_uuid
+        -- if there is no leader, we won't change the table
+        if leader.uuid ~= nil then
+            vars.leader_uuid = leader.uuid
+        end
         membership.set_payload('leader_uuid', vars.leader_uuid)
     end
+    vars.cache.is_leader = vars.leader_uuid == vars.instance_uuid
```
sergepetrenko, line 78: Is it possible that `vars.instance_uuid` is `nil`? At some point during or shortly after node startup, perhaps? If yes, you would set `is_leader` to true here when there is actually no leader.

test/integration/raft_test.lua
```diff
@@ -312,8 +312,8 @@ g.test_kill_master = function()
     t.assert_equals(res.status, 200)
     t.assert_equals(res.json, {})
 
-    start_server('storage-3')
     kill_server('storage-2')
+    start_server('storage-3')
```
sergepetrenko, line 316: Why is this change necessary?

sergepetrenko (commented):
Thanks for working on this!

Two questions from me.

cartridge/failover/raft.lua
yngvar-antonsson: No, it's not possible. `vars.instance_uuid` is set in `failover.cfg` before this trigger will be enabled.

test/integration/raft_test.lua
yngvar-antonsson: I think we discussed it before on call. Changing the positions of kill/start will reduce noise from raft and probably make the test more stable. But it's not a necessary change and I can remove it. What do you think?

cartridge/failover/raft.lua
sergepetrenko: Ok, thanks!

test/integration/raft_test.lua
sergepetrenko: But storage-3 is 'voter', it can't interfere with elections. I think this shouldn't affect the test stability, unless I forgot something else we discussed.

Let's remove this change.

yngvar-antonsson: Done.

sergepetrenko (approved):
Thanks!

PR #2210, opened 2024-04-01
Add more strict validation in `cartridge.is_healthy`

### LINKED ISSUE #2207: Missing some healthy RPC candidates in "healthy" cluster.
Sometimes tests fail due to missing some servers in a list returned from `rpc.get_candidates` call. Looks like this issue is due to health check inconsistency between `cluster_is_heathy` and `member_is_healthy` functions. The latter is more strict and checks for `state_prev` and `suspect` state.
cluster_is_healthy: 
```lua
local member = membership.get_member(server.uri) or {}                                      

if (member.status ~= 'alive') then                                                          
    return nil, . . .                                                                                        
elseif (member.payload.uuid ~= instance_uuid) then                                          
    return nil, . . .                                                                                       
elseif member.payload.state ~= 'ConfiguringRoles' and member.payload.state ~= 'RolesConfigured' then                                          
    return nil, . . .                                                     
end 
return true
```

member_is_healthy:
 ``` lua
   local member = membership.get_member(uri)                                                                                                                     
    return (                                                                                        
        (member ~= nil)                                                                             
        and (member.status == 'alive' or member.status == 'suspect')                                
        and (member.payload.uuid == instance_uuid)                                                  
        and (                                                                                       
            member.payload.state_prev == nil or -- for backward compatibility with old versions     
            member.payload.state_prev == 'RolesConfigured' or                                       
            member.payload.state_prev == 'ConfiguringRoles'                                         
        )                                                                                           
        and (                                                                                       
            member.payload.state == 'ConfiguringRoles' or                                           
            member.payload.state == 'RolesConfigured'                                               
        )                                                                                           
    )   
```

The issue is reproducible if some members have `ConfiguringRoles` as current state and `BoxConfigured` as prev state.

yngvar-antonsson:
Close #2207


psergee (approved):
Thank you for the patch! lgtm!

PR #2237, opened 2024-05-15
Use vshard-ee if available

yngvar-antonsson:


cartridge/lua-api/boxinfo.lua
```diff
@@ -51,8 +51,20 @@ local function get_info(uri)
...
+            vshard_version = vshard._VERSION .. ' EE'
+        elseif package.loaded['vshard'] ~= nil then
+            vshard = package.loaded.vshard
+            vshard_version = vshard._VERSION
+        end
 
-        local vshard = package.loaded.vshard
+        if package.loaded['ddl-ee'] ~= nil then
+            ddl_version = require('ddl-ee.version') .. ' EE'
+        elseif package.loaded['vshard'] ~= nil then
```
DifferentialOrange, line 65:
```suggestion
        elseif package.loaded['ddl'] ~= nil then
```

cartridge/lua-api/boxinfo.lua
```diff
@@ -51,8 +51,20 @@ local function get_info(uri)
         local membership_myself = require('membership').myself()
         local membership_options = require('membership.options')
 
+        local vshard, vshard_version, ddl_version
+        if package.loaded['vshard-ee'] ~= nil then
```
DifferentialOrange, line 55: Why not require here?

yngvar-antonsson: package.loaded was used in the previous version of this code. I can change it to require if necessary.

DifferentialOrange: I think it would be safer to use `require`, especially in case of ddl version extraction

yngvar-antonsson: I found out the reason why package.loaded was used. In some tests vshard might be disabled. Rare case, but I'd prefer to leave vshard with package.loaded. 
But I'll change package.loaded['ddl'] to package.loaded['ddl.version']

DifferentialOrange: `package.loaded['module.version']` and `require('module.version')` are antipatterns. Public API is `require('module')._VERSION`, and `require('module.version')` is just an internal implementation of this API (even though I don't remember any modules which use some other implementation).

Isn't `require('vshard')` should always be a success for regular cartridge since it is a rock dependency?

yngvar-antonsson: Fixed

PR #2269, opened 2024-08-15
Use membership instead of cache in disk_failure issues

### LINKED ISSUE #2265: Use membership in disk failure issues
Now issues have a cache with disk failure issues, it's not convenient and you can't get rid of the issue until restart. We should use membership payload to track the state of thous condition.

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #2265

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -37,6 +37,8 @@ Changed
 
 - uuids in issues replaces with instance names and uris.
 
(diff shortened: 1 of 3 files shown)


yngvar-antonsson:
This fix leads to unstable behavior. Closed.

PR #2339, opened 2025-06-25
log: add logging to failover module

vakhov:
What has been done:
Added structured logging to failover.lua and coordinator.lua. Now logs cover leader switches (with aliases and URIs), quorum and health checks, immunity status, and errors during failover decisions.

Why:
Previously, the failover process lacked visibility. It was hard to trace leader election logic or investigate why a leader was changed without deep manual inspection.

What problem is being solved:
This change improves observability of failover and leader appointment workflows. It makes it easier to debug production incidents and analyze unexpected leader transitions.

I didn't forget about

- [ ] Tests
- [x] Changelog
- [ ] Documentation

Close TNTP-3341

### DIFF
--- a/cartridge/failover.lua
+++ b/cartridge/failover.lua
@@ -217,7 +217,7 @@ end
 local function _get_replicaset_alias_by_replicaset_uuid(server_uuid)
     local topology_cfg = vars.clusterwide_config:get_readonly('topology')
 
-    local replicaset = topology_cfg.replicasets[server_uuid]
+    local replicaset = topology_cfg.replicasets[replicaset_uuid]
     if replicaset ~= nil and replicaset.alias ~= nil then
         return replicaset.alias
     end
--- a/cartridge/failover.lua
+++ b/cartridge/failover.lua
@@ -214,7 +214,7 @@ local function _get_appointments_stateful_mode(client, timeout)
     return client:longpoll(timeout)
 end
 
-local function _get_replicaset_alias_by_replicaset_uuid(server_uuid)
+local function _get_replicaset_alias_by_replicaset_uuid(replicaset_uuid)
     local topology_cfg = vars.clusterwide_config:get_readonly('topology')
 
     local replicaset = topology_cfg.replicasets[replicaset_uuid]
--- a/cartridge/failover.lua
+++ b/cartridge/failover.lua
@@ -214,6 +214,21 @@ local function _get_appointments_stateful_mode(client, timeout)
     return client:longpoll(timeout)
 end
 
+local function _get_replicaset_alias_by_server_uuid(server_uuid)
+    local topology_cfg = vars.clusterwide_config:get_readonly('topology')
+
+    local server = topology_cfg.servers[server_uuid]
+    if server then
+        local replicaset_uuid = server.replicaset_uuid
+        if replicaset_uuid then
+            local replicaset = topology_cfg.replicasets[replicaset_uuid]
+            if replicaset and replicaset.alias then
+                return replicaset.alias
+            end
+        end
+    end
+end
+
 local function describe(uuid)
     local topology_cfg = vars.clusterwide_config:get_readonly('topology')
     local servers = assert(topology_cfg.servers)
@@ -266,8 +281,10 @@ local function accept_appointments(appointments)
 
         changed = true
 
-        log.info('Replicaset %s%s: new leader %s, was %s',
+        local replicaset_alias = _get_replicaset_alias_by_server_uuid(leader_uuid) or ''
+        log.info('Replicaset %s(%s)%s: new leader %s, was %s',
             replicaset_uuid,
+            replicaset_alias,
             replicaset_uuid == vars.replicaset_uuid and ' (me)' or '',
             describe(leader_uuid),
             describe(current_leader)
--- a/cartridge/failover.lua
+++ b/cartridge/failover.lua
@@ -239,6 +239,8 @@ end
 -- @treturn boolean Whether leadership map has changed
 local function accept_appointments(appointments)
     checks('table')
+    log.debug('accept_appointments: incoming appointments = %s', json.encode(appointments))
+    local t_accept_appointments = fiber.clock()
     local topology_cfg = vars.clusterwide_config:get_readonly('topology')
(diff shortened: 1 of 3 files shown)


Satbek:
There is also roles/coordinator.lua where locates coordinator's logic.
I think we should add the reason for appointment in log message

https://github.com/tarantool/cartridge/blob/763de81461d48ac5d6aa219ca52f0f4ff8f28a6b/cartridge/roles/coordinator.lua#L106C18-L109C18

Satbek (changes requested):
Thanks for logging improvements.

I think we should log here an instance alias too

https://github.com/tarantool/cartridge/pull/2339/files#diff-b3b13e336d0031b78b83b2caa6fff8da5ce604376fecbe1c83723d0ab263a522L269-L274

Satbek (changes requested):
We also need to log replicaset's alias. It is in topology_cfg so we can add it.

About coordinator:

We need to log why an appointment was made and why it wasn’t.

**Log levels**

* `info` — when an appointment **is** made
* `warn` — when an appointment **is not** made

**When an appointment is made (`info`):**

* The current leader is unhealthy; after examining *N* candidates a new one was chosen.
* This is the first appointment; the first node in the topology was selected.

**When an appointment is not made (`warn`):**

* The replica set’s immunity timeout hasn’t expired.
* The current leader is alive and electable (may be too frequent to log; absence of such logs can itself imply everything is fine).

Suggested flow:

1. First check if the current leader is healthy.

   * If **not**, log the problem and proceed.
   * If **yes**, log that fact and exit.
2. If the leader is unhealthy but still under an immunity timeout, log the timeout and the decision not to switch.

Logs must make it clear:

* How candidates are evaluated
* Which replica sets were considered
* Why the master was changed or kept unchanged.

cartridge/roles/coordinator.lua
```diff
@@ -76,86 +87,121 @@ local function describe(uuid)
     return uuid
 end
 
+--- Make leader election decision for a replicaset.
+--
+-- This function evaluates the current leader and available candidates,
+-- taking into account immunity timeout and health status.
+--
+-- @function make_decision
+-- @tparam table ctx The failover context table (must include .members and .decisions)
```
Satbek, line 96:
```suggestion
-- @function make_decision
-- @local
-- @tparam table ctx The failover context table (must include .members and .decisions)
```

cartridge/roles/coordinator.lua
```diff
@@ -76,86 +87,121 @@ local function describe(uuid)
...
     while true do
-        log.info('control_loop: started')
+        log.info('control_loop: making decisions')
 
         ctx.members = membership.members()
 
         local updates = {}
 
-        for replicaset_uuid, _ in pairs(vars.topology_cfg.replicasets) do
+        for replicaset_uuid, data in pairs(vars.topology_cfg.replicasets) do
```
Satbek, line 178:
```suggestion
        for replicaset_uuid, replicaset in pairs(vars.topology_cfg.replicasets) do
```

Satbek (changes requested):
Hi,

```lua
local control_fiber = fiber.new(control_loop, session)
control_fiber:name('failover-coordinate')
```

Since the fiber is already named, that name will appear in the logs, so an extra prefix isn’t necessary.

The same applies to `fencing_healthcheck`:

```lua
vars.fencing_fiber = fiber.new(fencing_watch)
vars.fencing_fiber:name('cartridge.fencing')
```

The fiber name itself makes it clear where the log entry came from.

Let’s also keep the log line format:

```
Replicaset %s: appoint %s (%q) (manual)
```

so it always begins with “Replicaset …”.
For manual appointments, it would be helpful to include the replicaset alias as well.

Thanks!

vakhov:
Here is an example of the log output with the current changes applied:
````log
Making decisions
make_decision: evaluating replicaset 11111111-2222-3333-4444-555555555555 (candidates=3)
Replicaset 11111111-2222-3333-4444-555555555555(storage-replicaset): no appointment made (reason=immunity_not_expired, checked=0)
Wait membership notifications

Making decisions
make_decision: evaluating replicaset 11111111-2222-3333-4444-555555555555 (candidates=3)
Replicaset 11111111-2222-3333-4444-555555555555(storage-replicaset): no appointment made (reason=no_healthy_candidates, checked=3)
Wait membership notifications

Making decisions
make_decision: evaluating replicaset 11111111-2222-3333-4444-555555555555 (candidates=3)
Replicaset 11111111-2222-3333-4444-555555555555(storage-replicaset): appoint new leader bbbbbbbb-bbbb-cccc-dddd-eeeeeeeeeeee (storage-2, "localhost:3302"), previous leader aaaaaaaa-aaaa-bbbb-cccc-dddddddddddd (storage-1, "localhost:3301") (reason=new_leader_selected, checked=2)
Wait membership notifications
```

cartridge/failover.lua
```diff
@@ -214,18 +214,12 @@ local function _get_appointments_stateful_mode(client, timeout)
     return client:longpoll(timeout)
 end
 
-local function _get_replicaset_alias_by_server_uuid(server_uuid)
+local function _get_replicaset_alias_by_replicaset_uuid(server_uuid)
```
Satbek, line 217:
```suggestion
local function _get_replicaset_alias_by_replicaset_uuid(replicaset_uuid)
```

cartridge/failover.lua
```diff
@@ -214,18 +214,12 @@ local function _get_appointments_stateful_mode(client, timeout)
...
-    local server = topology_cfg.servers[server_uuid]
-    if server then
-        local replicaset_uuid = server.replicaset_uuid
-        if replicaset_uuid then
-            local replicaset = topology_cfg.replicasets[replicaset_uuid]
-            if replicaset and replicaset.alias then
-                return replicaset.alias
-            end
-        end
+    local replicaset = topology_cfg.replicasets[server_uuid]
```
Satbek, line 220:
```suggestion
    local replicaset = topology_cfg.replicasets[replicaset_uuid]
```

Satbek (commented):
LGTM

Add changelog please

Satbek (approved):

PR #2345, opened 2025-07-08
etcd-client: linearizable reads + round-robin endpoints fix stale leader map

### LINKED ISSUE #2342: Split-brain when etcd connectivity is fragmented: routers write to different masters after failover
**Failover mode :** stateful with an etcd v3 backend
**Replication** : synchronous 

**Problem description**

Under certain network partitions a router keeps an outdated leader map, continues to write to the former master, while other routers have already switched to the new master elected by the etcd quorum. As a result two masters accept writes concurrently → split-brain and data loss risk.

**How to reproduce**
1. Topology
      - one replicaset: 3 storages (storage-1, storage-2, storage-3)
      - 2 routers (router-1, router-2)
      - 3-node etcd cluster (etcd-A, etcd-B, etcd-C)
      - synchronous replication enabled
2. Create a network partition
    - Break traffic between storage-1 + router-1 and two of the etcd nodes, leaving them connected only to etcd-A.
    - Then break traffic between etcd-A and the remaining quorum (etcd-B, etcd-C).
    Result: router-1 and current master storage-1 see just one etcd node that has lost quorum.
3. Trigger master switch (from the healthy side):
    ```lua
         membership.events.generate('storage-1-uri', membership.options.DEAD)
    ```
    The majority side (storage-2, storage-2, router-2reaching the quorum) promotes a new master.
4. Observe
    - synchronous queue on storage-1 is abandoned, but it keeps read_only = false;
    - router-1 that lost quorum still sees storage-1 as leader and send writes there;
    - router-2 with quorum switch to storage-2 or storage-3
     Two independent masters are now serving writes.
(The same scenario without synchronous replication has not yet been checked.)

### Behaviour after storage / router regain access to the etcd quorum

Once **`storage-1` and `router-1` can reach `etcd-B`/`etcd-C` again**, the cluster still does **not converge**:

* `storage-1` keeps `read_only = false` and retains the abandoned synchronous queue (it still thinks it is the leader).
* `router-1` continues to read the stale leader map and sends writes to `storage-1`.
* `router-2` (and the other storages) still direct writes to `storage-2` (or `storage-3`).

So two masters remain active indefinitely; manual intervention is required to restore a single writable leader.

**Expected behaviour**
Routers (and storages) should never proceed with a leader map that cannot be confirmed by the etcd quorum; they should either converge to the same master or refuse writes until they can obtain a consistent view.

**Actual behaviour**

Leader map is read from a stale, minority etcd node, so different routers disagree on the master and split-brain occurs.

**Possible directions / open questions**

1. Quorum (linearizable) reads for active_leaders.
   Always fetch leader map with etcd linearizable read to avoid stale values held by a partitioned minority node.
2. Add network partition tests
 Cover scenarios with asymmetric connectivity and verify that all instances either agree on a single master or disable writes.
3. Router self-protection
    If a router cannot obtain the leader map with quorum it should temporarily disable itself (vshard.router.disable()) to prevent writes to an obsolete master.
4. Storage fencing fallback
    A storage that loses etcd connectivity during fencing should immediately switch to read_only = true, regardless of replica connectivity, until it regains quorum.

Satbek:
Fix for TNT-1357.

* Linearizable reads by default
   All ordinary reads to etcd (no wait) are now sent with quorum=true, guaranteeing the value comes from a majority node or fails fast.
* Strict round-robin over endpoints

    Every request (reads, writes and long-polls) advances the internal cursor and is issued to the next endpoint in the list.
    A single request can therefore no longer monopolise the same minority node.
* Long-polls (wait=true) unchanged except for rotation
    quorum=true is not added to wait requests (etcd rejects that combination).
    They still benefit from the new round-robin cursor: when the current long-poll completes, the next one goes to a different endpoint, ensuring we eventually reach a majority node.

- [x] Tests
- [x] Changelog
- [ ] Documentation

Close #2342

### DIFF
--- a/cartridge/etcd2.lua
+++ b/cartridge/etcd2.lua
@@ -68,6 +68,8 @@ local function request(connection, method, path, args, opts)
     assert(num_endpoints > 0)
     for _ = 1, num_endpoints do
         local eidx = connection.eidx
+
+        -- round-robin on etcd-connections
         connection.eidx = (connection.eidx % num_endpoints) + 1
 
         if #connection.endpoints ~= num_endpoints then


cartridge/etcd2.lua
```diff
@@ -29,6 +29,15 @@ local function request(connection, method, path, args, opts)
     assert(connection.etcd_cluster_id ~= nil)
 
     local body = {}
+    if method == 'GET' then
+        args = args or {}
+        -- Quorum does not work with wait, it does not wait result and returns immediately.
+        -- If quorum not provided explicitly by deafult it true.
+        if args.wait == nil and args.quorum == nil then
+            args.quorum = true
+        end
```
vakhov, line 38: Just a thought - maybe we could log when `quorum=true` is added implicitly (even at debug level)? Could be handy if someone’s chasing etcd inconsistencies.

Or at least worth mentioning somewhere in the docs - it's easy to forget it’s added by default.

a1div0 (approved):

cartridge/etcd2.lua
```diff
@@ -57,12 +66,9 @@ local function request(connection, method, path, args, opts)
...
     assert(num_endpoints > 0)
-
-    for i = 0, num_endpoints - 1 do
-        local eidx = connection.eidx + i
-        if eidx > num_endpoints then
-            eidx = eidx % num_endpoints
-        end
+    for _ = 1, num_endpoints do
+        local eidx = connection.eidx
+        connection.eidx = (connection.eidx % num_endpoints) + 1
```
vakhov, line 71: Just a nit: it took me a second to realize we’re using `eidx` first and then advancing `connection.eidx` for the next try. Not wrong at all, but maybe worth a tiny comment?

test/integration/etcd2_client_test.lua
```diff
@@ -578,3 +578,72 @@ function g.test_promote_after_close()
...
+
+function g.test_round_robin_on_etcd_nodes()
+    local client = create_client()
+    local session = client:get_session()
+    local eidx1 = session.connection.eidx
+    session.connection:request('GET', '/some_key', {quorum=false})
+    local eidx2 = session.connection.eidx
+    t.assert_not_equals(eidx1, eidx2)
+end
+function g.test_longpolling_lack_of_quorum()
```
vakhov, line 617:
```suggestion
end

function g.test_longpolling_lack_of_quorum()
```

test/integration/etcd2_client_test.lua
```diff
@@ -578,3 +578,72 @@ function g.test_promote_after_close()
...
+    t.assert_equals(resp, nil)
+    t.assert_equals(err.etcd_code, 100)
+
+    client:drop_session()
+    session = client:get_session()
+
+    resp, err = session.connection:request('GET', '/some_key', {quorum=true})
+    t.assert_equals(resp, nil)
+    t.assert_equals(err.etcd_code, nil)
+    t.assert_equals(err.http_code, 408)
```
vakhov, line 598: Could be interesting to add a case where one etcd node returns a quorum error (300/301), but another is healthy - just to check that we actually retry and succeed on the next node.

vakhov (approved):
Nice work! The round-robin + quorum fix looks solid and well-tested . Left a few minor comments around readability and a possible extra test case for quorum fallback. Nothing blocking from my side.

test/integration/etcd2_client_test.lua
vakhov: Got it — mocking etcd to reproduce 300/301 makes sense but not critical here. Thanks for the clarification! All good from my side 👍

olegrok (approved):

PR #2376, opened 2026-04-08
failover: fallback to first leader for new replicaset appointments

### LINKED ISSUE #2375: Deadlock in failover.cfg() when adding a new replicaset with stateful failover and enable_synchro_mode
**Failover mode:** stateful (`etcd2` / `stateboard`)  
**Replication:** synchronous (`enable_synchro_mode: true`)  
**Versions:** Cartridge `2.16.7`, Tarantool `2.11.8`

## Problem description

When a new replicaset is added to a running cluster via `edit_topology` (e.g. adding `shard-3` with two instances), the leader of the new replicaset gets stuck with `read_only=true` indefinitely. The `confapplier` never reaches `RolesConfigured`, the `twophase_commit` times out, and the entire `edit_topology` operation fails.

The issue is intermittent - it depends on whether the failover coordinator has discovered the new instances via membership and written the leader appointment to the state provider before failover.cfg() runs on the new instance.

## How to reproduce

1. Deploy a cluster with stateful failover (`etcd2` or `stateboard`), `enable_synchro_mode: true`, and at least one existing storage replicaset (e.g. `shard-1`, `shard-2`).
2. Start two new instances (`storage-3-1`, `storage-3-2`) and add them as a brand new replicaset `shard-3` via `edit_topology`.
3. The operation may time out with:

   ```text
   Failed to edit topology: "10.0.4.37:3311": timed out
   ```

4. In `storage-3-1.log`, the leader instance shows:

   ```text
   set 'read_only' configuration option to true
   ...
   Failover triggered, reapply scheduled (fiber 165)
   ...
   ER_READONLY: Can't modify data on a read-only instance - box.cfg.read_only is true
   ```

The bug reproduces roughly 1 in 3-5 runs.

## Root cause

There is a race condition in `failover.cfg()` (`failover.lua`) that leads to a deadlock:

1. `twophase.commit_2pc()` calls `confapplier.apply_config()` on the new instance.
2. `confapplier.apply_config()` sets state to `ConfiguringRoles` and calls `failover.cfg()`.
3. `failover.cfg()` fetches initial appointments from the state provider (`_get_appointments_stateful_mode`). Since the coordinator hasn't discovered the new replicaset via membership yet, there is no appointment for this replicaset - `first_appointments[replicaset_uuid] == nil`.
4. `accept_appointments()` + `constitute_oneself()`: the condition `active_leaders[replicaset_uuid] ~= instance_uuid` evaluates to `nil ~= uuid -> true`, so the instance considers itself not a leader and sets `is_rw = false`.
5. `box.cfg({read_only = true})` is applied.
6. `confapplier` then calls `roles.apply_config()`, which hangs because one or more roles try to write data on a `read_only` instance.
7. The `confapplier` is now stuck in `ConfiguringRoles` forever - it never transitions to `RolesConfigured` or `OperationError`.
8. Meanwhile, the failover coordinator eventually discovers the new instances and writes the leader appointment to the state provider. The failover loop fires `reconfigure_all()`.
9. Deadlock: `reconfigure_all()` calls `confapplier.wish_state('RolesConfigured', math.huge)` - it waits forever for `RolesConfigured`, but `confapplier` is stuck in step 7.

```text
confapplier: ConfiguringRoles
    -> failover.cfg() sets read_only=true (no leader in state provider)
    -> roles.apply_config() hangs on write in read_only mode
    -> stuck in ConfiguringRoles forever

failover loop: reconfigure_all()
    -> wish_state('RolesConfigured', math.huge)
    -> waits forever for confapplier to exit ConfiguringRoles
    -> never gets to set read_only=false

-> DEADLOCK
```

## Timeline from logs (real example)

```text
14:49:41.729  confapplier: state -> ConfiguringRoles
14:49:41.734  failover.cfg(): box.cfg({read_only = true})   <- no leader in etcd yet
14:49:41.735  roles.apply_config() starts
14:49:41.760  coordinator writes leader to etcd (21ms too late)
14:49:41.760  "Failover triggered, reapply scheduled"
14:49:41.760  reconfigure_all() -> wish_state('RolesConfigured', math.huge)  <- DEADLOCK
14:49:42.700  ER_READONLY on replication attempts
...
14:50:41.xxx  twophase_commit times out
```
## Expected behavior

When a new replicaset is created and its leader is not yet known to the state provider, `failover.cfg()` should default to the first instance from `failover_priority` (which is what the coordinator would assign anyway). This prevents `read_only=true` from being set on the actual leader.

## Actual behavior

`failover.cfg()` treats a missing appointment as "I am not the leader" and sets `read_only=true`, which deadlocks the entire configuration pipeline.

## Proposed fix

In `failover.cfg()`, inside the stateful mode branch, after fetching `first_appointments` from the state provider - if the current replicaset has no appointment, default to the first instance from `topology.get_leaders_order()`.

This is safe because:
- For a brand new replicaset, the coordinator will always assign `failover_priority[1]` as the leader.
  - Code reference: [`cartridge/roles/coordinator.lua`](https://github.com/tarantool/cartridge/blob/master/cartridge/roles/coordinator.lua#L146-L150) :
    ```lua
    local candidates = topology.get_leaders_order(vars.topology_cfg, replicaset_uuid)
    if current_decision == nil then
        -- This is a case when new replicaset is created.
        -- First appointment is always made according to `topology_cfg`
        -- without regard to the healthcheck
        local decision = pack_decision(candidates[1])
        ...
    end
    ```
- Once the coordinator writes the actual appointment, the failover loop will pick it up and correct if needed.

vakhov:
Fix a race in stateful failover when adding a brand new replicaset.

If the state provider has no appointment yet for the new replicaset, failover.cfg() now defaults to topology.get_leaders_order(...)[1] (failover_priority[1]) for initial appointments. This prevents the future leader from being switched to read_only before the coordinator publishes the first appointment and avoids deadlock during twophase apply.

- [x] Tests
- [x] Changelog
- [ ] Documentation

Close #2375

### DIFF
--- a/cartridge/failover.lua
+++ b/cartridge/failover.lua
@@ -980,6 +980,23 @@ local function cfg(clusterwide_config, opts)
             first_appointments = {}
         else
             first_appointments = appointments
+
+            -- gh-2375: for a brand new replicaset the state provider may not
+            -- have appointment yet. Fallback to failover_priority[1] to avoid
+            -- read_only deadlock during initial apply.
+            if first_appointments[vars.replicaset_uuid] == nil then
+                local leaders_order = topology.get_leaders_order(
+                    topology_cfg, vars.replicaset_uuid, nil, {only_enabled = true}
+                )
+                if leaders_order[1] ~= nil then
+                    first_appointments[vars.replicaset_uuid] = leaders_order[1]
+                    log.warn(
+                        'No leader for replicaset %q in state provider,' ..
+                        ' defaulting to %q from failover_priority',
+                        vars.replicaset_uuid, leaders_order[1]
+                    )
+                end
+            end
         end
 
         vars.failover_fiber = fiber.new(failover_loop, {
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,18 @@ and this project adheres to
 Unreleased
 -------------------------------------------------------------------------------
 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Fixed
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Fixed a stateful failover race when adding a brand new replicaset: if the
+  state provider has no appointment yet for this replicaset,
+  ``failover.cfg()`` now falls back to
+  ``topology.get_leaders_order(...)[1]`` (``failover_priority[1]``) for initial
+  appointment. This prevents the future leader from being switched to
+  ``read_only=true`` and avoids deadlock during topology apply (`#2375
+  <https://github.com/tarantool/cartridge/issues/2375>`_).
+
 -------------------------------------------------------------------------------
 [2.16.7] - 2026-03-20
 -------------------------------------------------------------------------------
--- a/test/integration/failover_stateful_add_replicaset_test.lua
+++ b/test/integration/failover_stateful_add_replicaset_test.lua
@@ -1,0 +1,295 @@
+local fio = require('fio')
+local t = require('luatest')
+local helpers = require('test.helper')
+
+local etcd2_client = require('cartridge.etcd2-client')
+local stateboard_client = require('cartridge.stateboard-client')
+
+local g_etcd2 = t.group('integration.failover_stateful.etcd2_add_replicaset')
+local g_stateboard = t.group('integration.failover_stateful.stateboard_add_replicaset')
+
+local router_uuid = helpers.uuid('a')
+local router_1_uuid = helpers.uuid('a', 'a', 1)
+
+local storage1_uuid = helpers.uuid('b')
+local storage1_1_uuid = helpers.uuid('b', 'b', 1)
+local storage1_2_uuid = helpers.uuid('b', 'b', 2)
+
+local storage2_uuid = helpers.uuid('c')
+local storage2_1_uuid = helpers.uuid('c', 'c', 1)
+local storage2_2_uuid = helpers.uuid('c', 'c', 2)
+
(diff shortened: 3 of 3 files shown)


cartridge/failover.lua
```diff
@@ -980,6 +980,23 @@ local function cfg(clusterwide_config, opts)
...
+            -- have appointment yet. Fallback to failover_priority[1] to avoid
+            -- read_only deadlock during initial apply.
+            if first_appointments[vars.replicaset_uuid] == nil then
+                local leaders_order = topology.get_leaders_order(
+                    topology_cfg, vars.replicaset_uuid, nil, {only_enabled = true}
+                )
+                if leaders_order[1] ~= nil then
+                    first_appointments[vars.replicaset_uuid] = leaders_order[1]
+                    log.warn(
+                        'No leader for replicaset %s in state provider,' ..
```
olegrok: `%s` -> `%q`

olegrok (dismissed):

cartridge/failover.lua
vakhov: Done

cartridge/failover.lua
```diff
@@ -980,6 +980,23 @@ local function cfg(clusterwide_config, opts)
...
+            -- read_only deadlock during initial apply.
+            if first_appointments[vars.replicaset_uuid] == nil then
+                local leaders_order = topology.get_leaders_order(
+                    topology_cfg, vars.replicaset_uuid, nil, {only_enabled = true}
+                )
+                if leaders_order[1] ~= nil then
+                    first_appointments[vars.replicaset_uuid] = leaders_order[1]
+                    log.warn(
+                        'No leader for replicaset %q in state provider,' ..
+                        ' defaulting to %s from failover_priority',
```
olegrok: The same (%q) was expected also here

olegrok (dismissed):

cartridge/failover.lua
vakhov: Done

olegrok (approved):

ita-sammann (approved):

Satbek (approved):

PR #2383, opened 2026-07-09
Retry box.ctl.promote() on conflict

ita-sammann:
When `box.ctl.promote()` returns ER_INTERFERING_PROMOTE during failover retry promotion 3 times with 1 second delay so the new master will not be stuck in read-only mode.

### DIFF
--- a/cartridge/failover.lua
+++ b/cartridge/failover.lua
@@ -454,27 +454,53 @@ local function fencing_start()
     vars.fencing_fiber:name('cartridge.fencing')
 end
 
+local SYNCHRO_PROMOTE_RETRY_ATTEMPTS = 3
+local SYNCHRO_PROMOTE_RETRY_DELAY = 1
+
 local function synchro_promote()
-    if vars.enable_synchro_mode == true
-    and vars.mode == 'stateful'
-    and vars.promote_for_leader_required
-    and vars.cache.is_leader
-    and not vars.failover_paused
-    and not vars.failover_suppressed
-    and box.ctl.promote ~= nil
-    then
+    local ok, err
+    for attempt = 1, SYNCHRO_PROMOTE_RETRY_ATTEMPTS do
+        if vars.enable_synchro_mode ~= true
+        or vars.mode ~= 'stateful'
+        or not vars.promote_for_leader_required
+        or not vars.cache.is_leader
+        or vars.failover_paused
+        or vars.failover_suppressed
+        or box.ctl.promote == nil
+        then
+            return
+        end
+
         log.info('Attempting box.ctl.promote()')
 
-        local ok, err = pcall(box.ctl.promote)
-        if ok ~= true then
-            log.error('Failed to promote: %s', err or 'unknown')
-            return err
+        ok, err = pcall(box.ctl.promote)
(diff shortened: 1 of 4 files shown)


cartridge/failover.lua
```diff
@@ -467,8 +472,23 @@ local function synchro_promote()
...
+                fiber.sleep(SYNCHRO_PROMOTE_RETRY_DELAY)
+                attempt = attempt + 1
+
+                ok, err = pcall(fiber.testcancel)
+                if ok ~= true then
+                    log.error('Fiber was cancelled in synchro_promote')
+                    return err
+                end
+
+                goto retry
```
Satbek: may be just `for i = 1, SYNCHRO_PROMOTE_RETRY_ATTEMPTS do` ?

cartridge/test-helpers/cluster.lua
```diff
@@ -349,8 +349,9 @@ function Cluster:join_server(server)
 end
 
 --- Blocks fiber until `cartridge.is_healthy()` returns true on main_server.
-function Cluster:wait_until_healthy(server)
-    self:retrying({}, function ()
+function Cluster:wait_until_healthy(server, config)
```
Satbek: may be rename `config` to `retry_opts` or `opts`

I thought that config is `cluster config`

test/integration/failover_synchro_test.lua
```diff
@@ -261,6 +261,91 @@ local function promote_errors_test(g)
...
+        for i, uuid in pairs({storage_1_uuid, storage_2_uuid, storage_3_uuid}) do
+            res = g.cluster:server(find_alias_by_uuid(uuid)):exec(function()
+                return {
+                    ro = box.info.ro,
+                    ro_reason = box.info.ro_reason,
+                    synchro = box.info.synchro,
+                }
+            end)
+
+            -- Storage 2 is RW, other storages are RO
```
Satbek, line 347: may be add some additional checks? with injections 

something like this:

https://github.com/tarantool/cartridge/blob/9adbb657e687a3c616c8ba193cffc69ec7880ce9/test/integration/manual_election_mode_helper_test.lua#L95-L113

to check that promote was called multiple times and box.error.INTERFERING_PROMOTE occurred

Satbek (commented):
Thanks! Great work!

some minor comments

p0rtale (dismissed):

cartridge/failover.lua
ita-sammann: Got rid of goto, added normal for loop

cartridge/test-helpers/cluster.lua
ita-sammann: Thanks for noticing! Renamed `config` to `opts`

test/integration/failover_synchro_test.lua
ita-sammann: That's a great idea, thanks! Added injection and check that promote is called 2 times and there is only one error and it is a 'instance X was promoted first' one.

p0rtale (approved):

Satbek (approved):

PR #2388, opened 2026-09-04
helpers: make vshard bootstrap timeout configurable

### LINKED ISSUE #2362: vshard: make bootstrap timeout configurable
In some cases a default hard-coded timeout is not enough to finish vshard bootstrap process:

https://github.com/tarantool/cartridge/blob/a875eb56e3a46079e3f483b26e8a4373100ed4f0/cartridge/roles/vshard-router.lua#L217

It will be great to make this option configurable.

patapenka-alexey:
The default vshard bootstrap timeout of 10 seconds may be not enough to
distribute buckets across the storages on slow CI runners.

vshard-router role:
- new option --vshard-bootstrap-timeout (env TARANTOOL_VSHARD_BOOTSTRAP_TIMEOUT)
  to configure the per-group vshard bootstrap timeout; the default value
  remains 10 seconds.

Test helpers (cartridge.test-helpers.Cluster):
- start(), bootstrap() and bootstrap_vshard() now accept an optional timeout
  that is forwarded to the bootstrap HTTP request. If it is omitted, the
  behavior is unchanged (no HTTP timeout is set). This allows tests to wait
  longer for the bootstrap to complete, e.g. cluster:start(60).

Also Server:graphql() no longer discards the caller-provided http options
when auth is enabled, so the timeout survives on auth-enabled clusters.

Part of TNTP-9875

### DIFF
--- a/cartridge/roles/vshard-router.lua
+++ b/cartridge/roles/vshard-router.lua
@@ -35,6 +35,7 @@ vars:new('routers', {
 
 vars:new('issues', {})
 vars:new('enable_alerting', false)
+vars:new('bootstrap_timeout', 10)
 
 -- Human readable router name for logging
 -- Isn't exposed in public API
@@ -80,9 +81,17 @@ local function init(_)
     local opts, _ = require('cartridge.argparse').get_opts({
         connections_limit = 'number',
         add_vshard_router_alerts_to_issues = 'boolean',
+        vshard_bootstrap_timeout = 'number',
     })
     if opts.add_vshard_router_alerts_to_issues ~= nil then
         vars.enable_alerting = opts.add_vshard_router_alerts_to_issues
+    end
+    if opts.vshard_bootstrap_timeout ~= nil then
+        local timeout = opts.vshard_bootstrap_timeout
+        if timeout <= 0 or timeout == math.huge then
+            error("vshard_bootstrap_timeout must be a finite positive number greater than 0", 0)
+        end
+        vars.bootstrap_timeout = timeout
     end
     local limit = opts.connections_limit
     if limit == nil then
@@ -214,7 +223,7 @@ local function bootstrap_group(group_name, vsgroup)
 
     log.info('Bootstrapping %s ...', router_name)
 
-    local ok, err = get(group_name):bootstrap({timeout=10})
+    local ok, err = get(group_name):bootstrap({timeout=vars.bootstrap_timeout})
     if not ok and err.code ~= vshard.error.code.NON_EMPTY then
         return nil, e_bootstrap_vshard:new(
             '%s (%s, %s)',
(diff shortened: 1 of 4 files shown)


patapenka-alexey:
Required for https://github.com/tarantool/graphqlapi-helpers/pull/56

cartridge/test-helpers/cluster.lua
```diff
@@ -276,7 +290,7 @@ function Cluster:start()
             self:wait_until_healthy(server)
         end
     else
-        self:bootstrap()
+        self:bootstrap(180)
```
vakhov: тут HTTP timeout выставлен в те же 180 секунд, что и серверный timeout, но это разные временные ограничения. здесь на весь запрос, а вишардовый на отдельные последовательные вызовы к репликасетам. получается что может быть такая ситуация что сервер ещё может продолжать bootstrap, а клиент уже завершит Cluster:start() по timeout. Кажется, для клиентского timeout стоит добавить запас либо считать оба ограничения от общего deadline.

cartridge/test-helpers/cluster.lua
```diff
@@ -237,7 +237,9 @@ function Cluster:apply_zone_distances()
 end
 
 -- Configure replicasets and bootstrap vshard if required.
```
vakhov:
```suggestion
--- Configure replicasets and bootstrap vshard if required.
```

cartridge/test-helpers/cluster.lua
vakhov: the client timeout is equal to the server timeout while the server may perform several sequential call with that timeout. as a result the client may time out before bootstrap finishes. may be would add some margin to the client timeout.

cartridge/roles/vshard-router.lua
```diff
@@ -214,7 +214,7 @@ local function bootstrap_group(group_name, vsgroup)
 
     log.info('Bootstrapping %s ...', router_name)
 
-    local ok, err = get(group_name):bootstrap({timeout=10})
+    local ok, err = get(group_name):bootstrap({timeout=180})
```
vakhov: this timeout applies to each callrw separately not to the entire bootstrap. as a result a request may wait for 180 seconds several times when a storage is unavailable. may be better to make this value configurable.

vakhov (commented):
Thanks for the patch! I left a few questions.

cartridge/test-helpers/cluster.lua
patapenka-alexey: Fixed.

cartridge/test-helpers/cluster.lua
patapenka-alexey: Replaced with a configurable timeout.

cartridge/roles/vshard-router.lua
patapenka-alexey: Replaced with a configurable timeout.

olegrok (approved):

Satbek (approved):

Satbek (approved):

vakhov (approved):

